PR #158, opened 2013-03-03
Quicker load and no debug print statement

benjaoming:
Thanks for a very useful project!! :)

I've removed an annoying print statement in `auto_discover()` regarding registered models. I don't find it useful since it's very easy to debug whether a model is registered or not, and I think the app/developer can live without it.

The catch-reraise pattern on non-`ImportError` has also been reduced to avoid first raising an ImportError and then detecting if there is a sub-module (which there obviously isn't when it's an ImportError).

### DIFF
--- a/modeltranslation/models.py
+++ b/modeltranslation/models.py
@@ -14,7 +14,7 @@ def autodiscover():
     from django.utils.importlib import import_module
     from django.utils.module_loading import module_has_submodule
     from modeltranslation.translator import translator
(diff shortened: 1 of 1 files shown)


last-partizan:
I'm closing this old pull-request.

benjaoming:
That's great, I think it was implemented in 72e2724 anyways -- but it happened back when Github didn't have notifications for that sort of stuff, so I wouldn't have known :)

PR #195, opened 2013-08-01
Better language selection in admin, inlines not tested.

ikresoft:
Better language selection, inlines not tested, and correction of pull request #190 to handle sr-latn code language.


modeltranslation/admin.py
```diff
@@ -1,7 +1,12 @@
 # -*- coding: utf-8 -*-
 from copy import deepcopy
+<<<<<<< HEAD
+from django.conf import settings
+from django.utils.translation import ugettext_lazy as _
+=======
```
matthiask: What are those merge markers doing here?

PR #239, opened 2014-04-02
Add the original content value (first tab content), in other tabs.

francescarpi:


fabiocaccamo:
Nice idea, but I think that it could be a little bit confusing to see all tabs already populated by default...
What do you think about:

- make it optional: MODELTRANSLATION_ADMIN_SHOW_DEFAULT_LANG_VALUE = True/False in 
settings.py

- default value shown as field placeholder or the help_text or with a different style

last-partizan:
Closing due to inactivity.

PR #273, opened 2014-10-20
Changes manager patching

mauler:
Doing this way if you choose to inherit MultilingualQuerySet
you application will works correctly with Pickle serialization.


last-partizan:
Closing due to inactivity.

PR #276, opened 2014-10-24
Sync/update_translation_fields verbosity and app selection, bis

wrwrwr:
This is a cleaned version of #274, please see the description there. Also including a somewhat different approach to filtering models by app that works with abstract models (not necessary here, but handy in upcoming `test_update_command` fixes).

PR #277, opened 2014-10-25
Fixes for test_update_command

wrwrwr:
Note: this is a follow-up to #276, the first 5 commits are exactly the same. The fixes here do depend on the enhancement proposed there, but in case you don't like the command app-selection feature, with some more work it should be possible to work around it.

I've been seeing various failures of the `test_update_command` for some time already, however none of them can be replicated with an empty project. They turned out not to be related to the command itself; rather the test acts as a beacon of side-effects and other quirks as its processing goes over all registered models. The failures fall into 3 categories:
- losing references to classes in `modeltranslation.tests.models` other than models, due to the hard-core reloading during the test base setup; usually apparent as some unexpected `None` during update of one of the custom manager models, for example ([M + MT](https://github.com/wrwrwr/mezzanine/compare/feature/modeltranslation-integration-1.7) + [1.7](https://github.com/django/django/commits/stable/1.7.x)):

```
File "/modeltranslation/tests/models.py", line 266, in get_queryset
    sup = super(CustomManager, self)
TypeError: must be type, not None

File "/modeltranslation/tests/models.py", line 294, in get_queryset
    return CustomQuerySet(self.model, using=self._db)
TypeError: 'NoneType' object is not callable
```
- Update not being effective, due to the command module being loaded before modeltranslation tests setup and sticking to some project's default language ([M + MT](https://github.com/wrwrwr/mezzanine/compare/feature/modeltranslation-integration-1.7) + [1.6](https://github.com/django/django/commits/stable/1.6.x)):

```
File "/modeltranslation/tests/tests.py", line 1903, in test_update_command
    self.assertEqual('initial', obj1.title_de)
AssertionError: 'initial' != u''
```
- Update trying to update user-models registered with project's languages different than the tests set (e.g. same as the last one plus this branch without 1aeaf69):

```
File "/django/db/models/sql/query.py", line 1283, in names_to_path
    "Choices are: %s" % (name, ", ".join(available)))
FieldError: Cannot resolve keyword 'something_de' into field. 
Choices are: <correct translation fields for project's language set>
```

In my experience, while working with the test code it's easy to run into other related issues, most perfidious are those caused by different parts of the system using different copies of `translator`. One hilarious was from an attempt to replace models reloading with lazy registration -- a `NotRegistered` error during `register`.

The idea is to avoid some module reloading in test preparation. The heavy changes are in the first two commits. First, in 9a74829, `reload_app` is asked to clean the translator registry from models that got reloaded, secondly 6c673cd refactors the test base to avoid reloading the translator module.

There is still some reloading left -- of `django.utils.translation` and `modeltranslation.settings`. The first one didn't happen to cause any trouble so far and as long as nothing imports or stores references to objects from modeltranslation settings the implemented module-level reference fixing should be general enough.

The only actual (at least as intended) test changes are in the last two commits: 383fa4b and 1aeaf69.

A nice thing is that with this branch you can run the Mezzanine tests together with modeltranslation suite without any failure, on Django 1.4 - 1.7, with or without South.

### DIFF
--- a/docs/modeltranslation/commands.rst
+++ b/docs/modeltranslation/commands.rst
@@ -8,6 +8,8 @@ Management Commands
 The ``update_translation_fields`` Command
 -----------------------------------------
 
+.. versionchanged:: 0.8
+
 In case modeltranslation was installed in an existing project and you
 have specified to translate fields of models which are already synced to the
 database, you have to update your database schema (see :ref:`db-fields`).
@@ -34,8 +36,9 @@ otherwise nothing is copied.
     the command will examine your ``settings.LANGUAGES`` variable and the first
     language declared there will be used as the default language.
 
-All translated models (as specified in the translation files) from all apps will be
-populated with initial data.
+On default, all translated models (as specified in the translation files) from
+all apps will be populated with initial data, but you can provide an ``--app``
+option to limit the population to models from a single app with a given label.
 
 
 .. _commands-sync_translation_fields:
@@ -44,6 +47,7 @@ The ``sync_translation_fields`` Command
 ---------------------------------------
 
 .. versionadded:: 0.4
+.. versionchanged:: 0.8
 
 .. code-block:: console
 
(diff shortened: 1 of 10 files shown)


last-partizan:
Closing due to inactivity.

PR #291, opened 2015-01-20
Changed MODELTRANSLATION_LANGUAGES example

pmourlanne:
If you use a tuple for this settings, you can get an error here : https://github.com/deschler/django-modeltranslation/blob/master/modeltranslation/utils.py#L48
`settings.AVAILABLE_LANGUAGES` is a tuple and cannot be concatenated with a list.

I guess it could be made clearer that you _have_ to use a list for this settings, or the problem lies in `utils.py`.

### DIFF
--- a/docs/modeltranslation/installation.rst
+++ b/docs/modeltranslation/installation.rst
@@ -195,7 +195,7 @@ Example::
         ('de', 'German'),
         ('pl', 'Polish'),
     )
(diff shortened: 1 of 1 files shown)


zlorf:
Already fixed in ef8e0b323fe2b9070b848e187b64d97d67396368

Issue #270, opened 2014-10-08
TabbedTranslationAdmin with no actionable fragment


helderco:
Currently with `TabbedTranslationAdmin`, when a tab is clicked, the screen scrolls to the top of the field, hiding the tabs themselves (assuming your form is bigger then your screen), because of the URL fragment (e.g. `#tab_id_description_pt`).

![tabbedtranslationadmin](https://cloud.githubusercontent.com/assets/174525/4559125/73e8be36-4ee8-11e4-8564-1a2c4c18edaa.png)

This is very annoying, especially if I want to quickly move into the other languages to check any missing or misspelled translations. I have to constantly scroll back up a bit to see the tabs.

It would be better if de fragment would target the top of the tabs instead of the field, but what would be ideal is to not scroll to anywhere at all, thus having the fragment not be actionable (empty return inside a click event in javascript).

fabiocaccamo:
The scroll is very annoying, it can be easily solved by using hashes different by tab ids, for example if the tab id is **#tab_id_name_en** the hash could be **#tab_name_en**.

In this way the work to do is to intercept the click, get the id, create the hash by replacing **#tab_id** with **#tab_**, then update the document location with the new hash.

fabiocaccamo:
@helderco @sunjoomoon @deschler I just fixed it and sent pull-request: https://github.com/deschler/django-modeltranslation/pull/297

shamuratov6563:
nice

PR #297, opened 2015-03-12
Fix bug #270

fabiocaccamo:
Add ```$.browser``` plugin import to avoid this error: ***Uncaught TypeError: Cannot read property 'msie' of undefined*** since ```$.browser``` was removed from jQuery starting with version 1.9. This error is the cause of page scrolling when changing language tab on an admin field. Tested on Django==1.6.5.


zlorf:
Where did ``Cannot read property 'msie' of undefined`` appear?

fabiocaccamo:
I discovered it using Google Chrome console. 
The error was thrown in jquery-ui.min.js at line 384.

fabiocaccamo:
Could you release a 0.8.1 version available through ```pip```?

zlorf:
I've asked @deschler to do so.

PR #298, opened 2015-04-01
Damn bug when using a queryset with select related

Satchitananda:
Please fix it and upload a fixed version on PyPI


zlorf:
@Satchitananda thank's for PR.
@deschler that's pretty nasty bug. Could you release 0.8.1? (I also merged another bugfix (probably))

PR #299, opened 2015-04-02
Use new meta API with django 1.8

lucacorti:
This seems enough to fix django 1.8 compatibility for me (deschler/django-modeltranslation/issues/295). Unfortunately I am not able to run tests, I added a 1.8 environment to tox.ini and also tried running runtests manually, but all tests involving the database fail. It seems syncdb is run after the tests.

```
----------------------------------------------------------------------
Ran 106 tests in 0.609s

FAILED (failures=1, errors=72)
Creating test database for alias 'default' (':memory:')...
Operations to perform:
  Synchronize unmigrated apps: modeltranslation
  Apply all migrations: (none)
Synchronizing apps without migrations:
  Creating tables...
    Running deferred SQL...
  Installing custom SQL...
Running migrations:
  No migrations to apply.
Destroying test database for alias 'default' (':memory:')...
(END) 
```

PR #302, opened 2015-04-12
Django 1.8 full compability

bsvetchine:
Fixes Django 1.8 compatibility without breaking backward compatibilities.


bsvetchine:
Duplicates with #299.

PR #313, opened 2015-05-11
Css improvements

fabiocaccamo:
- Changed admin-tabular-inlines vertical-align to middle (it was bottom, and it was very confusing for inlines containing big widgets like TinyMCE)
- Removed the annoying outline added to tab buttons by the browser
  schermata 2015-10-02 alle 18 02 19


fabiocaccamo:
What about this PR?

PR #317, opened 2015-05-21
Add mt-bidi class to input fields for bi-directional languages.

oliphunt:
Add a css class to distinguish between input fields for ltr and rtl languages, so that the text direction can be changed for all input fields that need to be bi-directional using a css rule:
```css
.mt-bidi {
    direction: rtl;
}
```


deschler:
Thank you!

PR #318, opened 2015-05-22
`translator.register` as a decorator

zenoamaro:
This patch adds a decorator version of `translator.register`, much like the new `admin.register` decorator introduced in Django 1.7, on which it is based.

The decorator is available in `decorators.py`, and also imported in `translator.py` for convenience during import. It accepts the same arguments as `register`. Tests and docs are of course provided.

~~~py
from modeltranslation.translator import register, TranslationOptions
from news.models import News

@register(News)
class NewsTranslationOptions(TranslationOptions):
    fields = ('title', 'text',)
~~~

### DIFF
--- a/docs/modeltranslation/registration.rst
+++ b/docs/modeltranslation/registration.rst
@@ -50,6 +50,18 @@ only imported. The ``NewsTranslationOptions`` derives from
 ``TranslationOptions`` and provides the ``fields`` attribute. Finally the model
 and its translation options are registered at the ``translator`` object.
 
+If you prefer, ``register`` is also available as a decorator, much like the
+one Django introduced for its admin in version 1.7. Usage is similar to the
+standard ``register``, just provide arguments as you normally would, except 
+the options class which will be the decorated one::
+
+    from modeltranslation.translator import register, TranslationOptions
+    from news.models import News
+
+    @register(News)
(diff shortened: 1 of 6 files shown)


zlorf:
Awesome patch! Docs updated and test included - thank you.

I've tried to do same thing, but with ``@translator.register(Model)`` and it failed as bare model can be registered as well and the situation was indistinguishable between decorator-syntax and normal-syntax. Your proposition to use brand new function - ``register`` is simple solution.

zlorf:
I think we can safely merge this.

deschler:
Confirmed, great enhancement. Thank you!

zenoamaro:
I don't know if there's a safe way to detect if a function has been called as a decorator, but this is more explicit and also what Django people are doing. I know magic usually ends with the explosion of the wizard's shack :).

I guess it could be more explicit still to have people import `register` from `decorators`, but I think this is a simple enough case that it can be exported along with other stuff without causing too much ambiguity.

zlorf:
No, there isn't any way.

Sadly, there can be a confusion: ``translator.register`` can be either function or decorator depending on whether it was imported like ``from modeltranslation import translator`` or ``from modeltranslation.translator import translator``. But let's hope nobody will have problems.

PR #321, opened 2015-07-01
Handle annotation fields when using values_list

lukaslundgren:
I recently tried to upgrade a project from v0.8b2 to v0.9.1 and ran into an issue regarding using annotate with values_list. The annotated values are lost.

The made up queryset `Foo.objects.values_list('field').annotate(Count('id'))` used to return `[('foo', 10), ...]` but as of v0.8 the returned value will be `[('foo', ), ...]`

This pull request contains one possible solution and a test case.


deschler:
Thank you!

PR #333, opened 2015-09-08
update_translation_fields: Support unmanaged models

akheron:
Add an option to update_translation_fields to skip unmanaged models. This is useful when models in an external database should be left untouched.


rmertens:
Is there any update on this? I want to use Modeltranslation with a database that is not managed by Django, so this seems very useful!

last-partizan:
Closing due to inactivity.

PR #360, opened 2016-02-03
let register decorator return decorated class

spacediver:
This way we preserve TranslationOptions subclass around in our module (and that's how decorators are supposed to work in the first place).

Thanks for magnificent app!


zlorf:
Thanks!

PR #368, opened 2016-03-23
Add primary support to DISTINCT statement

virgilio:


nealtodd:
+1 to merging support for using distinct(*fields) in the ORM.

An argumentless distinct() doesn't need any change , although it can produce correct but unexpected results because all fields, *including the language-suffixed ones produced by the modeltranslation migrations*, are used to determine distinctness (this might well be the reason for issue #334).

However, on PostgreSQL only, Django supports [passing field arguments to distinct()](https://docs.djangoproject.com/en/dev/ref/models/querysets/#distinct) to specify which fields the distinct should apply to (using the `DISTINCT ON` SQL expression).

This PR supports turning the field name into the correct field name for the current language in the expression. Without it, an ORM query like `manager.order_by('title').distinct('title')` will generate a SQL error because manager.order_by() rewrites the field name for the current language for the `ORDER BY` expression, but then does not match the (unrewritten) field name in the `DISTINCT ON` expression. 

Maybe not merged because there's no accompanying test?

The following test provides coverage for this PR. It'll only test the field name rewriting when the tests are running with a PostgreSQL database (e.g. using the `DB=postgres` environment var) since it's the only DB backend that supports `DISTINCT ON`. However, it still tests that a normal `DISTINCT` works as it should for all backends (as well as highlighting why the results are correct although can be unintuitive regarding duplicates).

Happy to get this test and the manager method into a single PR if it helps to get it accepted (or, @virgilio, feel free to add it directly to your PR).

```python
    def test_distinct(self):
        """Check that field names are rewritten in distinct keys."""
        manager = models.ManagerTestModel.objects
        manager.create(
            title_en='title_1_en', title_de='title_1_de',
            description_en='desc_1_en', description_de='desc_1_de')
        manager.create(
            title_en='title_1_en', title_de='title_1_de',
            description_en='desc_2_en', description_de='desc_1_de')
        manager.create(
            title_en='title_2_en', title_de='title_2_de',
            description_en='desc_1_en', description_de='desc_1_de')
        manager.create(
            title_en='title_2_en', title_de='title_2_de',
            description_en='desc_2_en', description_de='desc_2_en')

        # Without field arguments to distinct() all fields are used to determine
        # distinctness, therefore when only looking at a subset of fields in the
        # queryset it can appear that there are duplicates (the titles in this case)
        titles_for_en = tuple(m.title for m in manager.order_by('title').distinct())
        with override('de'):
            titles_for_de = tuple(m.title for m in manager.order_by('title').distinct())

        self.assertEqual(titles_for_en, ('title_1_en', 'title_1_en', 'title_2_en', 'title_2_en'))
        self.assertEqual(titles_for_de, ('title_1_de', 'title_1_de', 'title_2_de', 'title_2_de'))

        # On PostgreSQL only, distinct() can have field arguments (*fields) to specify which fields
        # the distinct applies to (this generates a DISTINCT ON (*fields) sql expression).
        # NB: DISTINCT ON expressions must be accompanied by an order_by() that starts with the
        # same fields in the same order
        if django_settings.DATABASES['default']['ENGINE'] == 'django.db.backends.postgresql_psycopg2':

            titles_for_en = tuple(
                (m.title, m.description) for m in manager.order_by('title', 'description').distinct('title'))
            with override('de'):
                titles_for_de = tuple(
                    (m.title, m.description) for m in manager.order_by('title', 'description').distinct('title'))

            self.assertEqual(titles_for_en, (('title_1_en', 'desc_1_en'), ('title_2_en', 'desc_1_en')))
            self.assertEqual(titles_for_de, (('title_1_de', 'desc_1_de'), ('title_2_de', 'desc_1_de')))
```

For reference, without the manager.distinct() method this PR adds, the unit test would error thusly:

```
ProgrammingError: SELECT DISTINCT ON expressions must match initial ORDER BY expressions
LINE 1: SELECT DISTINCT ON ("tests_managertestmodel"."title") "tests...

> print manager.order_by('title', 'description').distinct('title')).query

SELECT DISTINCT ON ("tests_managertestmodel"."title") "tests_managertestmodel"."id",
 "tests_managertestmodel"."title", "tests_managertestmodel"."title_de",
 "tests_managertestmodel"."title_en", "tests_managertestmodel"."visits",
 "tests_managertestmodel"."visits_de", "tests_managertestmodel"."visits_en",
 "tests_managertestmodel"."description", "tests_managertestmodel"."description_de",
 "tests_managertestmodel"."description_en" FROM "tests_managertestmodel" ORDER BY
 "tests_managertestmodel"."title_en" ASC, "tests_managertestmodel"."description_en" ASC

# Note the "title" in the DISTINCT ON expression and the "title_en" in the ORDER BY expression. 
```

deschler:
I'm going to add the test and merge the pull request. Since i don't use Postgres myself, i need someone to confirm it still works with current master after the merge.

PR #371, opened 2016-05-13
Works with collapsed admin elements

LeGast00n:
Tabs would not be initialized for fields inside a "collapsed" admin fieldset. This is meant to correct this.


deschler:
Looks straight forward. Thank you!

PR #378, opened 2016-06-27
Fix test failures when django directory is not writable

### LINKED ISSUE #377: Tests fails when django.contrib.auth.migrations is not writable
Hello,
I'm packaging modeltranslation for Debian and I would like to run the testsuite against the Django versions which is installed system-wide. Unfortunately, this does not work currently because it fails trying to create a migration:
```
ERROR: setUpClass (modeltranslation.tests.tests.FallbackTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/django/test/testcases.py", line 945, in setUpClass
    super(TestCase, cls).setUpClass()
  File "/home/rhertzog/deb/pkg/build-area/django-modeltranslation-0.11/modeltranslation/tests/tests.py", line 150, in setUpClass
    database=connections[DEFAULT_DB_ALIAS].alias)
  File "/usr/lib/python2.7/dist-packages/django/core/management/__init__.py", line 120, in call_command
    return command.execute(*args, **defaults)
  File "/usr/lib/python2.7/dist-packages/django/core/management/base.py", line 445, in execute
    output = self.handle(*args, **options)
  File "/usr/lib/python2.7/dist-packages/django/core/management/commands/makemigrations.py", line 143, in handle
    self.write_migration_files(changes)
  File "/usr/lib/python2.7/dist-packages/django/core/management/commands/makemigrations.py", line 172, in write_migration_files
    with open(writer.path, "wb") as fh:
IOError: [Errno 13] Permission denied: u'/usr/lib/python2.7/dist-packages/django/contrib/auth/migrations/0007_auto_20160627_2120.py'
```

I found a work-around by setting MIGRATION_MODULES to point migrations of the "auth" application to a package contained within modeltranslation and by copying into it the existing migrations from django.contrib.auth.migrations...

```
--- a/runtests.py
+++ b/runtests.py
@@ -40,6 +40,9 @@ def runtests():
                 'django.contrib.auth',
                 'modeltranslation',
             ),
+            MIGRATION_MODULES={
+                'auth': 'modeltranslation.tests.auth_migrations',
+            },
             ROOT_URLCONF=None,  # tests override urlconf, but it still needs to be defined
             LANGUAGES=(
                 ('en', 'English'),
```

rhertzog:
The tests modify the Group model exported by the django.contrib.auth
application and as such, when makemigrations is called, it wants to create
a new file in django/contrib/auth/migrations. This fails when the
directory is not writable, as is the case for a typical system-wide
installation (with a Debian package in my case).

To avoid this problem, we modify the test runner to set the
MIGRATION_MODULES django setting so that the migrations for the auth
application are looked up in the “modeltranslation.tests.auth_migrations”
package and we initialize that package with the migrations found
in the django directory.

Fix #377. This patch has been tested with Django 1.9.7.

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 import os
+import shutil
 import sys
 
 import django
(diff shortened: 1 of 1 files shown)


zlorf:
Fixed in the meantime; soon in the master branch.

PR #380, opened 2016-07-31
Feature: filtering and annotating querysets by translation status

maksimryndin:
Sometimes it is handy to get only translated instances in a specified language or be able to check translation status of each instance in the queryset.

Two methods for MultilingualManager (and MultilingualQuerySet) are proposed:
- `filter_translated()`
- `annotate_translated()`


last-partizan:
Closing due to inactivity.

PR #383, opened 2016-09-07
Make sync_translation_fields command Django 1.10 compatible

primoz-k:


zlorf:
Thanks you - however, this way the command wouldn't work in older Django versions. We need to maintain backward compatibility.

primoz-k:
How selfish of me, my bad :)

zlorf:
Thank you!

PR #384, opened 2016-09-14
Fix the Travis CI build

matthiask:


zlorf:
Brilliant PR! Many kudos for spotting problem with `travis.py` script.

PR #398, opened 2016-12-27
Fix order_by with expression

### LINKED ISSUE #375: MyModel.objects.order_by(Lower('name')) gives error
In Django 1.8 a new method Lower is created (see https://code.djangoproject.com/ticket/6498).

However, django-modeltranslation fails using it:

AttributeError
'Lower' object has no attribute 'startswith'

```
    Change translatable field names in an ``order_by`` argument
    to translation fields for the current language.
    """
    if not self._rewrite:
        return super(MultilingualQuerySet, self).order_by(*field_names)
    new_args = []
    for key in field_names:
                    **new_args.append(rewrite_order_lookup_key(self.model, key)) ...**
    return super(MultilingualQuerySet, self).order_by(*new_args)

def rewrite_order_lookup_key(model, lookup_key):
            if lookup_key.startswith('-'): ...
    return '-' + rewrite_lookup_key(model, lookup_key[1:])
else:
    return rewrite_lookup_key(model, lookup_key)
```

btoueg:
Fix #375 

 1. The fix could be moved within `rewrite_order_lookup_key` function
 2. I chose a try: except: approach but we could check if it's a string

### DIFF
--- a/modeltranslation/manager.py
+++ b/modeltranslation/manager.py
@@ -120,7 +120,15 @@ def append_lookup_key(model, lookup_key):
 
 
 def append_lookup_keys(model, fields):
-    return moves.reduce(set.union, (append_lookup_key(model, field) for field in fields), set())
+    new_fields = []
+    for field in fields:
+        try:
+            new_field = append_lookup_key(model, field)
+        except AttributeError:
+            new_field = (field,)
+        new_fields.append(new_field)
+
+    return moves.reduce(set.union, new_fields, set())
 
 
 def rewrite_order_lookup_key(model, lookup_key):
--- a/modeltranslation/manager.py
+++ b/modeltranslation/manager.py
@@ -351,7 +351,11 @@ class MultilingualQuerySet(models.query.QuerySet):
             return super(MultilingualQuerySet, self).order_by(*field_names)
         new_args = []
         for key in field_names:
-            new_args.append(rewrite_order_lookup_key(self.model, key))
+            try:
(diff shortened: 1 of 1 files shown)


btoueg:
If you want to check only without empty space : https://github.com/deschler/django-modeltranslation/pull/398/files?w=1

btoueg:
Tips : you can check the diff without the blanks here: https://github.com/deschler/django-modeltranslation/pull/398/files?w=1

btoueg:
CI failed because it couldn't install Pillow on Python 2.6

gabn88:
It would be really cool to have this merged in. 👍  It might need some additional review though, as it is a major change...

Currently all new Django ORM Goodies are not working when it is a translated model field. Hit my head against the rock again today.

raratiru:
Bumping on the rock!

raratiru:
I would like to ask if there is an anticipated date for getting in compliance with the new Django ORM. 

That would really save a couple of hits in the database, a fact that makes this compliance a very crucial issue.

btoueg:
New versions are being released without even reviewing this PR? @deschler 

I've rebased on top of 0.12.1

gabn88:
Bumping on the rock again, when using the qs1.union(qs2) method introducted in Django 1.11 in my DRF endpoint. If all the new features of Django 1.10+'s ORM stay unsupported in this module it is going to be time to migrate to another one I think.

btoueg:
Thx

PR #419, opened 2017-07-25
Fix reset of .select_related

tochev:
`.select_related` accepts `None` to clear any fields that might have been previously set.

Previously this resulted in the error `AttributeError: 'NoneType' object has no attribute 'split' `

This fix introduces an additional check.


deschler:
Thank you!

Issue #420, opened 2017-08-04
Incompatilbility with `django-audit-log` during tests


decibyte:
I have a Django project using both `django-audit-log` and `django-modeltranslation`.

I've been using the audit log for a while without any issues. Now that I introduced translations, a lot of my tests are blowing up. I'm sure the same would have happened if had introduced audit log after translations, so I'm not blaming any of the projects. I just wish they would both work together. But I'm unfortunately not able to find out what the problem is myself.

~~A weird thing is that this seems to only be a problem during testing. Everything seems to work as expected when running the project normally. But I want my tests to work as well.~~

I've created [another small example Django project](https://github.com/decibyte/django-auditlog-modeltranslation-testcase) to provide a case for this issue.

The complete output from running tests is here:

```
$ ./manage.py test
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
E
======================================================================
ERROR: test_my_model (app.tests.MyModelTestCase)
Test the model.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/path/to/django-auditlog-modeltranslation-testcase/app/tests.py", line 12, in setUp
    text='Text of Instance 1'
  File "/path/to/django-auditlog-modeltranslation-testcase/.env/lib/python3.5/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/path/to/django-auditlog-modeltranslation-testcase/.env/lib/python3.5/site-packages/modeltranslation/manager.py", line 381, in create
    return super(MultilingualQuerySet, self).create(**kwargs)
  File "/path/to/django-auditlog-modeltranslation-testcase/.env/lib/python3.5/site-packages/django/db/models/query.py", line 394, in create
    obj.save(force_insert=True, using=self.db)
  File "/path/to/django-auditlog-modeltranslation-testcase/.env/lib/python3.5/site-packages/django/db/models/base.py", line 807, in save
    force_update=force_update, update_fields=update_fields)
  File "/path/to/django-auditlog-modeltranslation-testcase/.env/lib/python3.5/site-packages/django/db/models/base.py", line 847, in save_base
    update_fields=update_fields, raw=raw, using=using,
  File "/path/to/django-auditlog-modeltranslation-testcase/.env/lib/python3.5/site-packages/django/dispatch/dispatcher.py", line 193, in send
    for receiver in self._live_receivers(sender)
[...]

Issue #200, opened 2013-09-26
Making unittest using models with translated fields


dnmellen:
Hi, I'm trying to make some unittests and I have a problem with the current version of django-modeltranslation.

I'm creating objects in test's `setUp()` but since this code runs outside a django view it's failing and giving me this exception:

`Category` model has some translated fields, `name` is one of them.

```
cat = Category.objects.create(catalog=c, name="Category one")

TypeError: 'name_es' is an invalid keyword argument for this function
```

I really need to create objects in test code, Is there some way to achieve this?

decibyte:
@dnmellen, I know it's a long time ago now, but did you happen to also be using `django-audit-log` at the same time? In that case (and for others coming here), I found out you also have to remember to register your audit log models for translation. As seen in [this example](https://github.com/decibyte/django-auditlog-modeltranslation-testcase/compare/fix?expand=1#diff-f5e2c57158f98362114fabca111c67dfR7).

ice3:
It's a really old issue, but I saw it while debugging the same issue. 
When you call the management command, don't forget to make the migration before running the tests.

PR #427, opened 2017-09-10
Add section about django-audit log

decibyte:
Improve the documentation by describing how to use `django-modeltranslation` in combination with `django-audit-log`.

Refs #420, #200.


deschler:
Thank you!

Issue #413, opened 2017-06-21
Issues with complex M2M relationships


Schnouki:
Hi,

I'm using [django-oscar](https://github.com/django-oscar/django-oscar) with django-modeltranslation, and I have a few issues with some classes in its [catalogue](http://docs.oscarcommerce.com/en/latest/ref/apps/catalogue.html) app. I'm translating the `name` and `description` fields in the `Category` class.

# Models and their relationships
Models hierarchy:

    MP_Node ---> AbstractCategory ---> Category

          AbstractProductCategory ---> ProductCategory

                  AbstractProduct ---> Product

`AbstractCategory` is an abstract model that inherits from `MP_Node`, another abstract model from [django-treebeard](https://django-treebeard.readthedocs.io/en/latest/mp_tree.html). This makes it possible to easily define hierarchies of categories, and extend Oscar's `AbstractCategory` model in the user's code. `MP_Node` uses a custom manager, and `Product` has two simple custom managers as well (defined in `AbstractProduct`).

The `Category` model is then linked to the `Product` model via the `ProductCategory` model:

    categories = models.ManyToManyField("catalogue.Category, through="ProductCategory")

Again, in Oscar they actually are defined as simple children of the `AbstractProduct` and `AbstractProductCategory` models.

# The issue
Now, in my code, I want to search products that have certain categories, by category name:

    products = Product.objects.filter(categories__name = "Foo")

This doesn't work, and when logging the SQL queries it's easy to see why:

    SELECT * FROM "catalogue_product" 
    INNER JOIN "catalogue_productcategory" ON ("catalogue_product"."id" = "catalogue_productcategory"."product_id")                                                                                                                                                          
    INNER JOIN "catalogue_category" ON ("catalogue_productcategory"."category_id" = "catalogue_category"."id")
    WHERE "catalogue_category"."name" = 'Foo'         

The `WHERE` clause does the search on `name` instead of `name_fr`... So no good results.

# Workaround
After reading some other issues, especially #389, I came up with a workaround:
- redefine the `Category` custom manager (`objects = MP_NodeManager()`, like in `MP_Node`)
- extend the managers from `AbstractProduct`, make them inherit from `MultilingualManager`, and add them to `Product`

Now everything works as expected and my tests pass :)

However, I think that:
- if this behavior is expected, it should be documented somewhere
- if it's not, well, it should work without needing further modification of unrelated models and managers.

[...]

Issue #389, opened 2016-10-19
Django 1.10 default manager not MultilingualManager for model inheriting abstract class with a manager defined on objects (leading to AttributeError: 'QuerySet' object has no attribute 'rewrite')


nealtodd:
Whilst upgrading my application to Django 1.10 I came across the following issue where the default manager for a Model Translation registered model didn't get dynamically set to `modeltranslation.manager.MultilingualManager`.

This resulted in an exception being thrown in [update_translation_fields.py L32](https://github.com/deschler/django-modeltranslation/blob/master/modeltranslation/management/commands/update_translation_fields.py#L32)

`AttributeError: 'QuerySet' object has no attribute 'rewrite'`

A similar issue was resolved in #123 but it seems something in the internal changes in Django 1.10 has resurfaced it in the particular scenario I have (not one covered by #381).

I haven't yet identified what the changes are that cause it but I've narrowed the scenario down to one in which a model inherits from an abstract model that explicitly sets a manager via the  `objects` attribute. In this case the resulting `_default_manager` is that set on `objects` rather than `modeltranslation.manager.MultilingualManager`.

The simplest example I can boil it down to is this:

```
# models.py

class FooAbstract(models.Model):
    class Meta:
        abstract = True

class Foo(FooAbstract):
    name = models.CharField(blank=True, max_length=1)

class Bar(models.Model):
    objects = models.Manager()
    name = models.CharField(blank=True, max_length=1)

class BazAbstract(models.Model):
    objects = models.Manager()
    class Meta:
        abstract = True

class Baz(BazAbstract):
    name = models.CharField(blank=True, max_length=1)


# translations.py

@register(Foo)
class FooTranslationOptions(TranslationOptions):
    fields = ('name',)

@register(Bar)
class BarTranslationOptions(TranslationOptions):
    fields = ('name',)

@register(Baz)
class BazTranslationOptions(TranslationOptions):
    fields = ('name',)


# shell

from modeltranslation.translator import translator
models = translator.get_registered_models(abstract=False)
for model in models:
    print(model.__name__, type(model._default_manager))

# Django 1.9 output:

Foo <class 'modeltranslation.manager.MultilingualManager'>
Bar <class 'modeltranslation.manager.MultilingualManager'>
Baz <class 'modeltranslation.manager.MultilingualManager'>

# Django 1.10 output:

Foo <class 'modeltranslation.manager.MultilingualManager'>
[...]

PR #431, opened 2017-09-26
Fix M2M and inherited managers

Schnouki:
This PR fixes 2 closely related issues:
- #389: models that inherit from an abstract model with a custom manager are not properly patched
- #413: ManyToMany fields defined in abstract models don't work as expected (because of the previous one)

This PR creates:
- models that exhibit the bug in #389 through one of the previously existing tests
- a test that reproduces #413 

Then there's the fix for both bugs :)

Tested with Django 1.10 and 1.11 (Python 3.6), and 1.8 and 1.9 (Python 2.7).

### DIFF
--- a/modeltranslation/tests/models.py
+++ b/modeltranslation/tests/models.py
@@ -351,3 +351,61 @@ class MultitableConflictModelA(models.Model):
 
 class MultitableConflictModelB(MultitableConflictModelA):
     title = models.CharField(ugettext_lazy('title'), max_length=255)
+
+
+# ######### Complex M2M with abstract classes and custom managers
+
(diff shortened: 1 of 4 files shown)


deschler:
Thanks guys, that's an important one! Merging it to master.

dmarcelino:
Hi @deschler! Are there any plans for releasing a new version of `django-modeltranslation` soon? This unreleased PR fixes the breaking tests for https://github.com/infoportugal/wagtail-modeltranslation/pull/150 and I'd prefer not adding a git commit to the package requirements.

Thanks and Merry Xmas!

PR #434, opened 2017-10-29
Add docstring documentation for utils.unique

bennylope:
Reading through the code I thought "this function is redundant!, just use `set`!". It appears not to be redundant however, and a short explanation of what it does, and why, should provide clarity when reading the module.


deschler:
Thank you!

PR #444, opened 2017-12-15
- fix distinct with *fields

simon-the-sloth:


modeltranslation/tests/tests.py
```diff
@@ -2540,6 +2540,25 @@ def test_order_by(self):
         self.assertEqual(titles_asc, ('a', 'b', 'c'))
         self.assertEqual(titles_desc, ('c', 'b', 'a'))
 
+    def test_distinct(self):
+        """
+        Check that field names are rewritten in distinct fields.
+        This test works only on postgres.
```
last-partizan, line 2546: Why this works only on postgres?

PR #459, opened 2018-03-19
Fix add new inline and handles themes with inlines in tabs:

helio-correia:
- Themes like django jet makes the inlines in a hidden tab, there you can add inlines
 - In these themes when you are on a tab and you click save and continue editing the other tabs looses the grouping translated tabs


helio-correia:
After a few tests later this gets some errors

PR #470, opened 2018-06-29
fix: copy form defined widget attrs

chubz:
Hi, 

I've noticed in my project that custom defined widget attrs are not used on translated fields in Admin.
This might be a naive fix, so if you have anything you would like to add I can try and add in or cover.

I think this problem might also apply for other things that can get defined in meta like:
help_texts dict or error_messages ...


chubz:
@deschler or @zlorf any input on this? Would it be a welcome addition as a minor fix for next release?

PR #476, opened 2018-09-11
Added renderer attribute for django 2.1 compatibility

orbitvu:
Django 2.1 removed support for Widget.render() methods without the renderer argument: https://docs.djangoproject.com/en/2.1/releases/2.1/#features-removed-in-2-1

### DIFF
--- a/modeltranslation/widgets.py
+++ b/modeltranslation/widgets.py
@@ -57,17 +57,20 @@ class ClearableWidgetWrapper(Widget):
         """
         return self.widget.media + self.checkbox.media + Media(self.Media)
 
-    def render(self, name, value, attrs=None):
+    def render(self, name, value, attrs=None, renderer=None):
         """
         Appends a checkbox for clearing the value (that is, setting the field
         with the ``empty_value``).
         """
-        wrapped = self.widget.render(name, value, attrs)
+        wrapped = self.widget.render(name, value, attrs, renderer)
         checkbox_name = self.clear_checkbox_name(name)
         checkbox_id = self.clear_checkbox_id(checkbox_name)
         checkbox_label = self.clear_checkbox_label
         checkbox = self.checkbox.render(
-            checkbox_name, value == self.empty_value, attrs={'id': checkbox_id})
+            checkbox_name,
+            value == self.empty_value,
+            attrs={'id': checkbox_id},
+            renderer=renderer)
         return mark_safe(self.template.format(
             conditional_escape(wrapped),
             conditional_escape(checkbox_id),


modeltranslation/widgets.py
```diff
@@ -57,17 +57,20 @@ def media(self):
...
+        wrapped = self.widget.render(name, value, attrs, renderer)
         checkbox_name = self.clear_checkbox_name(name)
         checkbox_id = self.clear_checkbox_id(checkbox_name)
         checkbox_label = self.clear_checkbox_label
         checkbox = self.checkbox.render(
-            checkbox_name, value == self.empty_value, attrs={'id': checkbox_id})
+            checkbox_name,
+            value == self.empty_value,
+            attrs={'id': checkbox_id},
+            renderer=renderer)
```
matthiask: This will break with older Django versions. Not sure what versions you're trying to support, but I'd use `**kwargs` instead for passing on the renderer argument (as long as you're not using the renderer instance anyway)

modeltranslation/widgets.py
```diff
@@ -57,17 +57,20 @@ def media(self):
         """
         return self.widget.media + self.checkbox.media + Media(self.Media)
 
-    def render(self, name, value, attrs=None):
+    def render(self, name, value, attrs=None, renderer=None):
```
zerolab:
```suggestion
    def render(self, name, value, attrs=None, *args, **kwargs):
```

modeltranslation/widgets.py
zerolab:
```suggestion
             *args,
            **kwargs)
```

modeltranslation/widgets.py
```diff
@@ -57,17 +57,20 @@ def media(self):
...
         """
         Appends a checkbox for clearing the value (that is, setting the field
         with the ``empty_value``).
         """
-        wrapped = self.widget.render(name, value, attrs)
+        wrapped = self.widget.render(name, value, attrs, renderer)
```
zerolab:
```suggestion
         wrapped = self.widget.render(name, value, attrs, *args, **kwargs)
```

modeltranslation/widgets.py
```diff
@@ -57,17 +57,20 @@ def media(self):
...
         checkbox_name = self.clear_checkbox_name(name)
         checkbox_id = self.clear_checkbox_id(checkbox_name)
         checkbox_label = self.clear_checkbox_label
         checkbox = self.checkbox.render(
-            checkbox_name, value == self.empty_value, attrs={'id': checkbox_id})
+            checkbox_name,
+            value == self.empty_value,
+            attrs={'id': checkbox_id},
```
zerolab:
```suggestion
             {'id': checkbox_id},
```

zerolab:
On further thought, this is not needed as the `renderer` argument was added in 1.11 and django-modeltranslation only supports Django 1.11+ in latest master.

Also, it would be great to add 2.1 to tox.ini and .travis.yml

PR #482, opened 2018-10-25
Changes on manager.py

Danlestal:
Manager file modified to catch the Exception on get_options_for_model in case the model is not registered for translation

### DIFF
--- a/modeltranslation/manager.py
+++ b/modeltranslation/manager.py
@@ -85,8 +85,13 @@ def append_fallback(model, fields):
     """
     fields = set(fields)
     trans = set()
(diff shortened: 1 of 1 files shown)


last-partizan:
Why is this required?

Danlestal:
If I am honest to you I cant remember why I made this modification. It was on my previous job some months ago, and unfortunately I do not have access to the source code anymore.

I recall that it was something related with the way migrations were being applied, but as I said, I can't remember it properly :/

last-partizan:
Probably it doesn't required anymore :)

Closing pull-request.

PR #485, opened 2018-11-09
Add app_label and model_name arguments to update_translation_fields command

fdemmer:
This updates the `update_translation_fields` command to satisfy a requirement I had in a project with proxy-models and especially *database views* using [django-pgviews](https://github.com/mypebble/django-pgviews). 

- While the command works just fine with proxy-models, updating fields on the proxy-model, that were already updated using the concrete model is redundant.
- When using pgviews, model classes are written with the `Meta` option `managed = False`. The model's fields are constructed (or reflected) to match the database view and can be used like a normal model incl. registering with `modeltranslation`, it is just not used with migrations. Obviously a database view is not writable, therefore the command fails; it should only operate on the tables/models the view is based on.

This changes the command so that proxy models (`proxy = True`) and models not managed by Django (`managed = False`) are generally ignored. Furthermore it adds two optional, positional arguments to update only fields in models of a specific app or even just a single model.

### DIFF
--- a/modeltranslation/management/commands/update_translation_fields.py
+++ b/modeltranslation/management/commands/update_translation_fields.py
@@ -15,10 +15,12 @@ class Command(BaseCommand):
         verbosity = options['verbosity']
         if verbosity > 0:
             self.stdout.write("Using default language: %s" % DEFAULT_LANGUAGE)
+
         models = translator.get_registered_models(abstract=False)
         for model in models:
             if verbosity > 0:
                 self.stdout.write("Updating data of model '%s'" % model)
+
             opts = translator.get_options_for_model(model)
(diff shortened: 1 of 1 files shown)


fdemmer:
uh, should have looked through previous pull requests more carefully; this one does almost the same as me: https://github.com/deschler/django-modeltranslation/pull/333

last-partizan:
Well, it looks good.

Do you still want to merge it or you found another workaround for problem?

fdemmer:
i used exactly this/my fork to work around this issue. i think it makes a useful addition and would appreciate it being merged, so i can switch back to using the upstream package!

last-partizan:
Merged, thank you.

PR #500, opened 2019-03-28
Fix fields list bug in admin.TranslationAdmin

### LINKED ISSUE #332: group_fieldsets leaves ManyToMany fields out
When using group_fieldsets, ManyToMany fields are not displayed in the admin anymore. I think this happens because ManyToMany fields are not listed in Model._meta.fields.

numrut:
Fixes #332


last-partizan:
Thanks, accepted.

PR #502, opened 2019-03-30
feat(make-jquery-urls-configurable): Add settings

mes3yd:
Add 3 new settings (without MODEL_TRANSLATION prefix) that deal with external resources.
This allows software with elevated security restrictions to function properly.


last-partizan:
Looks good, accepted.

PR #507, opened 2019-04-22
add fallback_languages_per_field to TranslationOptions

wtayyeb:
There is `model_fallback_languages` in `TranslationOptions` to define fallback languages for all fields.
but sometime we want to only fallback specific field and let other not to fallback or fallback on different languages. For example assume a model with title, description fields which we want to translate them for en, de but we want to fallback title to en when no translated version of it in de and don't fallback for description field. this is where `fallback_languages_per_field` will help us:

``` python
class MyModel(models.Model):
    title = models.CharField(...)
    description = models.TextField(...)
    ...

@register(MyModel)
class MyModelTranslationOptions(TranslationOptions):
    fields = ('title', 'description')
    fallback_languages_per_field = {
        'title': {'default': ('en', 'de'), },
    }
```


last-partizan:
Well, it looks like simple change.

And probably useful one. Do you still want to merge it or do you found another way to do it for your case?

If you want to merge, then add some documentation and tests for this feature.

wtayyeb:
Absolutely I think this is important for any real world multilingual projects. 

My workaround for all patches like this which are in waiting to merge state is to publish them as separate package with -pr000 suffix.
following link is for this PR and I use it since creation.
https://pypi.org/project/django-modeltranslation-pr507/

last-partizan:
Good, please add some docs here: https://github.com/deschler/django-modeltranslation/blob/master/docs/modeltranslation/registration.rst about this new option.

And tests, if you have idea about how to test it.

And the i'll merge it.

PR #509, opened 2019-04-28
Update installation doc

davideghz:
syncdb command is deprecated in django 1.7. Use the python manage.py migrate instead.


davideghz:
Hi, thx for merging it. Pls note the doc did not change at https://django-modeltranslation.readthedocs.io/en/latest/installation.html#setup

last-partizan:
Thanks for reporting, i'll see what can i do.

davideghz:
Hi! I think `python manage.py syncdb` has been removed since v 1.9 (see: https://docs.djangoproject.com/en/dev/releases/1.9/#features-removed-in-1-9). The doc is still incorrect if I'm not missing something.

Issue #510, opened 2019-05-24
Signature for `formfield_for_dbfield` needs updatng


mes3yd:
[Back in 2015](https://github.com/django/django/commit/dbb0df2a0ec5bee80bee336fc81408efb30b7e47), the signature of `formfield_for_dbfield` was changed. This is rather problematic now, since it's missing in the `super()` call and signature for model translation's version. Any attempt to override or extend this method, `patch_translation_field` or deal with formsets is met with difficulties or impossibilities.

last-partizan:
Fixed in master.

PR #513, opened 2019-07-02
Fixed 'formfield_for_dbfield' method signature. See #510

fedenko:
It's a fix for `TypeError: formfield_for_dbfield() takes 2 positional arguments but 3 were given` issue when you try to override `formfield_for_dbfield` method in `TranslationAdmin`. See #510.


last-partizan:
Looks good, accepting :)

fedenko:
Thanks! :smile:

PR #519, opened 2019-10-03
Fix admin exclude usage

dmydlo:
This updates the usage of `exclude` admin field. Current implementation skips `get_exclude` which can mess up with either multiple inheritance classes or custom setting of `exclude`.
The easiest error to generate with it is to create an admin like that:
```python
class OurModel(models.Model):
    name = models.CharField(_("name"), max_length=255, unique=True)
    description = models.TextField(_("description"))

@modeltranslation.translator.register(models.OurModel)
class OurModelTranslationOptions(modeltranslation.translator.TranslationOptions):
    fields = ("name", "description")
    required_languages = ("en",)

@admin.register(models.OurModel)
class OurModelAdmin(translation_admin.TranslationAdmin):
    exclude = []
```
then let's say we want to exclude `description` for instances whose `name` start with letter "A".
First, let's add two instances, one with a `name` starting with "A" and other not.
We edit Admin:
```python
@admin.register(models.OurModel)
class OurModelAdmin(translation_admin.TranslationAdmin):
    exclude = []

    def get_exclude(self, request, obj=None):
        if obj and obj.name.startswith("A"):
            return ["description"]
        else:
            return []
```
No effect whatsoever as `get_exclude` is not used. So let's just overwrite the `self.exclude`:
```python
@admin.register(models.OurModel)
class OurModelAdmin(translation_admin.TranslationAdmin):
    exclude = []

    def get_exclude(self, request, obj=None):
        if obj and obj.name.startswith("A"):
            self.exclude = ["description"]
        else:
            self.exclude = []
        return super().get_exclude(request, obj)
```
The `super().get_exclude()` probably will have one line `return self.exclude`.
We test it... and it workes?
Yes, until we choose to first get to the edit page of the instance starting with "A" and then to the other one.

> KeyError at /admin/category/ourmodel/1/change/
> "Key 'description_en' not found in 'OurModelForm'. Choices are: ...

What happened? `get_exclude` wasn't called. Fix? F5 everytime after seeing this error will do the job, as well as the proposed change :)

### DIFF
--- a/modeltranslation/admin.py
+++ b/modeltranslation/admin.py
@@ -184,12 +184,13 @@ class TranslationBaseModelAdmin(BaseModelAdmin):
         """
         Generic code shared by get_form and get_formset.
         """
-        if self.exclude is None:
+        exclude = self.get_exclude(request, obj)
+        if exclude is None:
             exclude = []
         else:
-            exclude = list(self.exclude)
+            exclude = list(exclude)
         exclude.extend(self.get_readonly_fields(request, obj))
-        if not self.exclude and hasattr(self.form, '_meta') and self.form._meta.exclude:
+        if not exclude and hasattr(self.form, '_meta') and self.form._meta.exclude:
             # Take the custom ModelForm's Meta.exclude into account only if the
             # ModelAdmin doesn't define its own.
             exclude.extend(self.form._meta.exclude)


last-partizan:
Thanks, merged it.

PR #520, opened 2019-10-05
Add django 3.0 to the test matrix

mvbrn:


last-partizan:
@mvbrn we can not merge this until it passes tests. Can you fix this?

last-partizan:
@mvbrn any news on this?

mvbrn:
Unfortunately, error is not so obvious and probably it's related to some changes in internal mechanics of Django ORM. Have no time to dig into it :/

last-partizan:
I've got some time to look into this. And it was already almost fixed, but hardcoded to use correct code only for django 2, not 3.

Thanks for your contribution :ok_hand:

Issue #523, opened 2019-10-20
Wrong default_manager when inheriting from an abstract model


deegeenl:
When creating a model with a custom manager (`objects`) that inherits from an abstract base class that also defines a custom manager (e.g. `translations`), the default manager of the concrete class gets substituted by the `translations` manager from the base class. 

The reason is that when patching the managers, the inherited managers get added to the local managers of the child models, but the counter isn't changed. Therefore, when fetching the default manager, Django picks the one with the lowest counter which is usually the one defined in the parent (abstract) class. Before the patching, the inherited manager has a higher `depth` in the inheritance chain, and therefore isn't considered potential default if the model has its own custom managers.

Here's an example:

```
class CustomQuerySet(models.query.QuerySet):
    pass


class CustomManager2(models.Manager):
    def get_queryset(self):
        return CustomQuerySet(self.model, using=self._db)
    get_query_set = get_queryset


class CustomManagerAbstract(models.Manager):
    def to_translate(self):
        return self.get_queryset().filter(needs_translation=True)


# Abstract base class with one extra (non-default) custom manager: `translations`
class CustomManagerBaseModel(models.Model):
    needs_translation = models.BooleanField(default=False)

    objects = models.Manager()
    translations = CustomManagerAbstract()

    class Meta:
        abstract = True


# Child class that redefines its default manager: `objects`
class CustomAbstractManagerTestModel(CustomManagerBaseModel):
    title = models.CharField(ugettext_lazy('title'), max_length=255)

    objects = CustomManager2()


@register([
    models.CustomAbstractManagerTestModel
])
class CustomManagerTestModelTranslationOptions(TranslationOptions):
    fields = ('title',)
```

And here's the test that doesn't pass:
```
    def test_default_manager_for_inherited_models(self):
        """Test if default manager is still set from local managers"""
        manager = models.CustomAbstractManagerTestModel._meta.default_manager
        self.assertEqual('objects', manager.name)
```

I've already created a patch that I will submit as pull request shortly.

PR #524, opened 2019-10-20
Fix for issue #523

deegeenl:
Make sure model managers added as local managers have their counter reset, so that they don't override the default manager. This fixes issue #523


last-partizan:
Thanks, merged it.

Issue #527, opened 2019-11-13
Tests are not working with django2.2 when django files are not writable.


keuko:
Hello guys,

We have test issues with python3-django-modeltranslation when packaging in debian with django2.2.

Build process steps:
  - Install all requirements
  - Run tests
  - If tests are OK, package is created.

Tests permissions problem description:
When package is building, it is expected  that no files are written outside directory where build process is running.
Problem is that tests in django-modeltranslation are trying to write migrations files to /usr/lib/python3/dist-packages/django/contrib/auth/migrations/ which is failing, because build process is running under NON-root user, while dependecies are installed under OS pythonpath with   root ownership.

When we were building previous versions of django, we were using this commit as patch https://github.com/deschler/django-modeltranslation/commit/f9d7390eeaf5b7b96aaeaee92bc259a1e9423fdb to make tests work when django files are not writable. Now this patch is not working with django 2.2.

You can reproduce issue by changing ownership to root ( or some other user ) in tox py37-2.2 env and run tests under different user. Applied patch is also not working.

keuko:
debian@debian-sid-builder:/tmp/django-modeltranslation$ tox -e py37-2.2
GLOB sdist-make: /tmp/django-modeltranslation/setup.py
py37-2.2 inst-nodeps: /tmp/django-modeltranslation/.tox/dist/django-modeltranslation-0.13.2.zip
py37-2.2 installed: Django==2.2.7,django-modeltranslation==0.13.2,Pillow==6.2.1,pkg-resources==0.0.0,pytz==2019.3,six==1.13.0,sqlparse==0.3.0
py37-2.2 runtests: PYTHONHASHSEED='1909180129'
py37-2.2 runtests: commands[0] | django-admin.py --version
2.2.7
py37-2.2 runtests: commands[1] | /tmp/django-modeltranslation/.tox/py37-2.2/bin/python runtests.py
Creating test database for alias 'default' ('file:memorydb_default?mode=memory&cache=shared')...
Operations to perform:
  Synchronize unmigrated apps: modeltranslation
  Apply all migrations: auth, contenttypes
Synchronizing apps without migrations:
  Creating tables...
    Running deferred SQL...
Running migrations:
  Applying contenttypes.0001_initial... OK
  Applying contenttypes.0002_remove_content_type_name... OK
  Applying auth.0001_initial... OK
  Applying auth.0002_alter_permission_name_max_length... OK
  Applying auth.0003_alter_user_email_max_length... OK
[...]

PR #528, opened 2019-11-13
Fix failing tests when local migrations are used (Fixes: #527).

keuko:
The following tests issues are adressed in this patch:

- Settings are changed so that local tests.auth_migrations are used
  instead of django.contrib.auth.migrations
- Distributed django.contrib.auth.migrations are copied to
  tests.auth_migrations for tests
- All migrations are deleted from tests.auth_migrations on teardown


last-partizan:
Thanks, merged.

Issue #382, opened 2016-08-12
Using `.full_clean` with fallback languages on instance before saving leads to updating incorrect fields.


sdeprez:
First of all, thanks for your work, amazing library! 

Otherwise, I'm facing the following issue. Let's say we have the model:

```
from django.db import models

class MyObject(models.Model):
    name = models.CharField(max_length=255, blank=True, null=True)
```

with the settings

```
LANGUAGES = [ 
    ('en', _('English')),
    ('fr', _('French')),
]
MODELTRANSLATION_FALLBACK_LANGUAGES = ('en', 'fr')
```

And we execute the following code:

```
my_object = MyObject.objects.create()

# Assign a value ONLY to name_fr
my_object.name_fr  = 'some_fr_name'

# Here is the line that causes the problem!
my_object.full_clean()
my_object.save()

#
my_object.refresh_from_db()
print my_object.name_en   # print 'some_fr_name'!
```

So the problem is that `name_en` is updated even if we only specify the value of `name_fr`! If you remove the French from fallback values, or if you don't do the full clean, it works fine. It also works fine if `name_en` had a value before.

zlorf:
Interesting, will take a look in following days.

rcatajar:
@zlorf @deschler Issue is still present on the latest version. 

translation.py
```python

class TestObjectTO(TranslationOptions):
    fields = ('test', )


translator.register(TestObject, TestObjectTO)
```

models.py
```python
class TestObject(models.Model):
    test = models.CharField(max_length=100)
```

settings.py
```python

LANGUAGE_CODE = 'en'
LANGUAGES = (
    ('en', _('English')),
    ('fr', _('French'))
)
MODELTRANSLATION_LANGUAGES = ('en', 'fr')
MODELTRANSLATION_DEFAULT_LANGUAGE = 'en'
MODELTRANSLATION_FALLBACK_LANGUAGES = ('en', 'fr')
```

test shell:
```python
In [2]: obj = TestObject.objects.create()

In [3]: obj.test_fr = "foobar"

In [4]: obj.test_en

In [5]: obj.full_clean()

In [6]: obj.test_en
Out[6]: 'foobar'

In [7]: obj.save()

In [8]: obj.refresh_from_db()

In [9]: obj.test_en
Out[9]: 'foobar'
```

Any idea what can be the root cause ? I'm not familiar with the source code, but I'm willing to take time to submit a PR if you can point me to the right direction.

buczek:
[...]

PR #536, opened 2020-01-16
Don't leak into current language from .clean_fields()

buczek:
Django uses .clean_fields() to verify the values of all fields of an
instance, e.g. after the instance has been updated from a form view.

.clean_fields() essentially walks over the fields in the model, reads
the current values, verifies and normalizes them, and writes them
back [1].

With model-translations activated, this can leak values from a fallback
language into the current language of the session, if the value for the
current language has been set to an empty value from a translation aware
application or form.

Example:

- current language is "en"
- fallback languages are "en", "de"
- web form wants to set "title_de" to "Schnapsidee" and leave
    "title_en" empty

After the form is submitted, Django calls .clean_fields() which
processes the field "title" first

- reads value of "title" - which gives "Schnapsidee" from
    "title_de" by the fallback mechanism of Modeltranslations
- writes this value back to "title", which sets "title_en"
    as a side effect.

Bugs have been reported at [2], [3] and [4]

Disable the implicit setting of the translation field from the
translated field ( Rule 2 of [5] ) during .clean_field() processing.

[1] https://github.com/django/django/blob/bf77669453b9e9f64291da6701fe06fd5ba47e29/django/db/models/base.py#L1229
[2] https://github.com/deschler/django-modeltranslation/issues/382
[3] https://github.com/infoportugal/wagtail-modeltranslation/issues/247
[4] https://github.molgen.mpg.de/molgen/mpicms/issues/15
[5] https://django-modeltranslation.readthedocs.io/en/latest/usage.html#rules-for-translated-field-access

### DIFF
--- a/modeltranslation/fields.py
+++ b/modeltranslation/fields.py
@@ -314,9 +314,10 @@ class TranslationFieldDescriptor(object):
         instance.__dict__[self.field.name] = value
         if isinstance(self.field, fields.related.ForeignKey):
             instance.__dict__[self.field.get_attname()] = None if value is None else value.pk
-        if getattr(instance, '_mt_init', False):
+        if getattr(instance, '_mt_init', False) or getattr(instance, '_mt_disable', False):
             # When assignment takes place in model instance constructor, don't set value.
             # This is essential for only/defer to work, but I think it's sensible anyway.
+            # Setting the localized field may also be disabled by setting _mt_disable.
             return
         loc_field_name = build_localized_fieldname(self.field.name, get_language())
         setattr(instance, loc_field_name, value)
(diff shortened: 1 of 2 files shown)


buczek:
**Maintainer**:
* Make the `_mt_disable` attribute counting to support recursion? There often is recursion, because the `patch__*` functions are called multiple times, when model classes and their base classes are patched. Not a problem here, though, because nothing relevant happens after `_mt_disable` is set to `False` to early by an inner invocation.
* Put the flag/counter somewhere else than into the instance?
* Get rid of `_mt_init` and use the same flag/counter (which should become a counter if there are multiple users) ?
* Any other approach preferred ?

last-partizan:
I think everything is fine.

Make flake8 happy, and i'll merge it :)

buczek:
> modeltranslation/fields.py:317:51: E271 multiple spaces after keyword

Picard, if you know what I mean.

PR #537, opened 2020-02-12
base_manager shouldnt be the same as default_manager

Sult:
…ty that creates a manager each time it is required. So overwriting didnt work. Now i am far from a metaclass expert and im sure this can be done neater, however this is all the time i have at the moment. Improvements shouldnt be too hard now the issue has a (ugly) fix. Also added a short test, but i am used to pytest, so forgive me for the ugly manager solution. Hope this can help you out.

I am happy to improve on any suggestions made before merging

### DIFF
--- a/modeltranslation/tests/tests.py
+++ b/modeltranslation/tests/tests.py
@@ -41,7 +41,7 @@ models = translation = None
 request = None
 
 # How many models are registered for tests.
-TEST_MODELS = 33 + (1 if MIGRATIONS else 0)
+TEST_MODELS = 35 + (1 if MIGRATIONS else 0)
 
 
 class reload_override_settings(override_settings):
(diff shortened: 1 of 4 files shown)


modeltranslation/tests/tests.py
```diff
@@ -990,6 +992,38 @@ def test_reverse_relations(self):
         self.assertIn(fk_inst_en, test_inst2.test_fks.all())
         self.assertNotIn(fk_inst_de, test_inst2.test_fks.all())
 
+    def test_reverse_lookup_with_filtered_queryset_manager(self):
```
last-partizan: This test should be failing with master branch, but it works. Are you sure you testing what you think you are testing?

modeltranslation/translator.py
```diff
@@ -124,6 +125,43 @@ def __str__(self):
         return '%s: %s + %s' % (self.__class__.__name__, local, inherited)
 
 
+class MultilingualOptions(options.Options):
+    """ Far from a MetaClass expert, maybe this could mbe combined with TranslationOptions"""
+
+    @cached_property
+    def base_manager(self):
```
last-partizan: Can we not copy-paste django code, but instead just call `super()`, check if Manager is not Multilingual and override __class__?

Sult: I found it a bit dirty myself as well, I think meanwhile i figured out a way that might work. Coming back to it

modeltranslation/tests/tests.py
Sult: Guess not. Ill have another look

last-partizan:
Now it looks fine enough. Merging.

PR #541, opened 2020-03-04
Update to use `gettext_lazy`

ThijsBoehme:
The use of `ugettext` and variants is deprecated in Django 3.0 and to be removed in Django 4.0.
See more here: https://code.djangoproject.com/ticket/30165


last-partizan:
Python 2 is dead, so i'm merging this and will remove python 2 from test matrix in the next release.

last-partizan:
But wait, there is other references to ugettext. Can you replace them all?

ThijsBoehme:
Absolutely. I did only notice these as they threw a warning in my project.

PR #549, opened 2020-04-21
Fix: prevent a new migration from being created each time makemigrations is run

gregoiredx:
When a translated field references a django auth model, a  new migration file is created each time `makemigrations` is run.

This is because modeltranslation generates a new manager `NewMultilingualManager` with a `deconstruct()` method (see [here](https://github.com/deschler/django-modeltranslation/blob/53d2fc9a9ca4d4488ee5a166fa6c5456cacd6f93/modeltranslation/translator.py#L189-L201)) but without a `__eq()__` method.
See [django doc on `deconstruct()`](https://docs.djangoproject.com/en/3.0/topics/migrations/#adding-a-deconstruct-method):
> To prevent a new migration from being created each time `makemigrations` is run, you should also add a `__eq__()` method to the decorated class. This function will be called by Django’s migration framework to detect changes between states.

So, in this PR, `NewMultilingualManager.__eq__()` is implemented in a way that prevent `makemigrations` from seeing it as a new manager.

When adding a `__eq__()` method, the default `__hash__()` method of the object is suppressed (see [here](https://docs.python.org/3.8/reference/datamodel.html#object.__hash__)).
We need `NewMultilingualManager` to be hashable since it's used as one (see [here](https://github.com/deschler/django-modeltranslation/blob/53d2fc9a9ca4d4488ee5a166fa6c5456cacd6f93/modeltranslation/translator.py#L221-L222) for example).
So `__hash__()` is also implemented here with the same implementation as in django [`BaseManager`](https://docs.djangoproject.com/en/2.2/_modules/django/db/models/manager/).

### DIFF
--- a/modeltranslation/tests/tests.py
+++ b/modeltranslation/tests/tests.py
@@ -4,6 +4,7 @@ from unittest import skipUnless
 import datetime
 import fnmatch
 import imp
+import io
 import os
 import shutil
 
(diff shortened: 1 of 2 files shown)


last-partizan:
Thank you, merged it.

gregoiredx:
So nice, thanks for your reactivity @last-partizan !
Could you make a new release with this fix?

last-partizan:
Yes, i'll do it later today :)

last-partizan:
Released :)

gregoiredx:
super, thanks again

PR #563, opened 2020-08-07
Adds a language option to the update_translation_fields commands

MrCordeiro:
Now, not only the _default language_ will have its translation field populated, but you can provide a ``--language`` option to populate any other language listed on your ``settings.py``.

**My use case was as follows:**

I needed to search an Object by its `name` field, like so:

```python
Object.objects.filter(name__icontains="query")
```

But when someone from Italy tried searching for a translatable object, it would look for the `name_it` field only. If that field was empty, it wouldn't search into the default `name` field as well. The most straightforward solution was to populate all translatable fields regardless of whether they were using the default language or not.

So, if I run:

```shell
python manage.py update_translation_fields --language it
``` 

All the empty `name_it` fields that are empty will contain the value from the `name` field.

### DIFF
--- a/modeltranslation/management/commands/update_translation_fields.py
+++ b/modeltranslation/management/commands/update_translation_fields.py
@@ -1,14 +1,17 @@
 # -*- coding: utf-8 -*-
 from django.db.models import F, Q
-from django.core.management.base import BaseCommand
+from django.core.management.base import BaseCommand, CommandError
 
-from modeltranslation.settings import DEFAULT_LANGUAGE
+from modeltranslation.settings import AVAILABLE_LANGUAGES, DEFAULT_LANGUAGE
 from modeltranslation.translator import translator
 from modeltranslation.utils import build_localized_fieldname
 
 
+COMMASPACE = ", "
+
+
 class Command(BaseCommand):
-    help = ('Updates empty values of default translation fields using'
+    help = ('Updates empty values of translation fields using'
             ' values from original fields (in all translated models).')
 
     def add_arguments(self, parser):
@@ -19,6 +22,12 @@ class Command(BaseCommand):
         parser.add_argument(
             'model_name', nargs='?',
             help='Model name to update empty values of only this model.',
+        )
+        parser.add_argument(
+            '--language',
+            action='store',
+            help=('Language translation field the be updated.'
+                  ' Default language field if not provided')
         )
 
     def handle(self, *args, **options):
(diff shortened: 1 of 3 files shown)


modeltranslation/management/commands/update_translation_fields.py
```diff
@@ -1,14 +1,17 @@
...
-from django.core.management.base import BaseCommand
+from django.core.management.base import BaseCommand, CommandError
 
-from modeltranslation.settings import DEFAULT_LANGUAGE
+from modeltranslation.settings import AVAILABLE_LANGUAGES, DEFAULT_LANGUAGE
 from modeltranslation.translator import translator
 from modeltranslation.utils import build_localized_fieldname
 
 
+COMMASPACE = ", "
```
last-partizan, line 10: This is defined, but not used?

pyproject.toml
```diff
@@ -1,18 +1,19 @@
...
+maintainers = ["Serhii Tereshchenko <serg.partizan@gmail.com>"]
+
+[tool.poetry.dependencies]
+python = "^3.8"
+django = "^3.0.4"
+
+[tool.poetry.dev-dependencies]
+pdbpp = "^0.10.2"
+flake8 = "^3.7.9"
+six = "^1.15.0"
```
last-partizan, line 15: Why are your files contains ^M characters?

https://stackoverflow.com/questions/1889559/git-diff-to-ignore-m

modeltranslation/management/commands/update_translation_fields.py
MrCordeiro: Oops! It's being used now. Line 58.

pyproject.toml
MrCordeiro: I have no idea... Maybe because I'm on Windows?
Maybe if I just edit the file on Github they stop showing?

MrCordeiro: Should I delete this file? 
I've only added six because the test suite wouldn't run otherwise, but it not integral to the PR.

last-partizan:
@MrCordeiro yes, you'd better delete this line.

If you install it with `poetry install`, it should be installed from setup.py as dependency along with django.

last-partizan:
@MrCordeiro as for advice on ^M - have you followed link to stackoverflow? https://stackoverflow.com/questions/1889559/git-diff-to-ignore-m

You should enable `git config --global core.autocrlf true`.

MrCordeiro:
> @MrCordeiro yes, you'd better delete this line.
> 
> If you install it with `poetry install`, it should be installed from setup.py as dependency along with django.

Line deleted, as it was merely ancillary.

(I had followed the link, but was unable to understand that stackoverflow discussion :-/ ...)

last-partizan:
Thanks, merged it!

Issue #565, opened 2020-08-18
Inline item, addition broken


VDShark:
Tabs don't applied to new inlines.
It worked in 0.14, but since 0.15 broken.

![image](https://user-images.githubusercontent.com/8749506/90495679-92cf1e00-e14d-11ea-9d4b-556dc6f164d2.png)

![image](https://user-images.githubusercontent.com/8749506/90495714-9b275900-e14d-11ea-9aa7-8d3351ab557f.png)

last-partizan:
Please check if it's working in 0.16.2

VDShark:
Yes, it works.
Thanks :)

PR #584, opened 2021-02-17
Fix/add another

21adrian1996:
The encapsulation into window onload will not work if the onload method is overwritten later by another package (for example it isn't working properly with django suits).

As document ready seems to be late enough for the method to work properly, I think simply removing it should be a good fix.


21adrian1996:
This realtes to https://github.com/deschler/django-modeltranslation/issues/565

last-partizan:
We're not using `window.onload`, we're using `$(window).on("load"`, are you sure it gets overriden by another packages?

You have tested this, and it's working for you?

21adrian1996:
Not sure if it is overwritten or otherwise not fired, but it does not work properly like this.

I have implemented this fix in one of our projects and it works perfectly also in combination with django-suits.

last-partizan:
Well, let's try and hope it doesn't break something :)

PR #592, opened 2021-04-15
fix .latest() ORM method with django 3.2

### LINKED ISSUE #591: `.latest()` ORM method is broken: Cannot filter a query once a slice has been taken.
The last release (0.17) of `django-modeltranslation` has broken the usage of [`.latest()`](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#latest) with django 3.2.

``` python
from django.db import models
from modeltranslation.translator import TranslationOptions, register

class TranslatedModel(models.Model):
    translated_field = models.CharField(max_length=255, null=True)

@register(models.TranslatedModel)
class TranslatedModelTranslationOptions(TranslationOptions):
    fields = ("translated_field",)
```

Then in a django shell:
``` bash
>>> from bug_app.models import TranslatedModel
>>> TranslatedModel.objects.latest("id")
...
AssertionError: Cannot filter a query once a slice has been taken.
```

MDziwny:
fix #591

### DIFF
--- a/modeltranslation/manager.py
+++ b/modeltranslation/manager.py
@@ -300,7 +300,7 @@ class MultilingualQuerySet(models.query.QuerySet):
     def _rewrite_filter_or_exclude(self, args, kwargs):
         if not self._rewrite:
             return args, kwargs
-        args = map(self._rewrite_q, args)
+        args = tuple(map(self._rewrite_q, args))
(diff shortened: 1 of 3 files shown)


last-partizan:
Whoa, nice fix :)

I'll merge it as soon as it passed tests :)

PR #594, opened 2021-04-17
Use github actions for CI

last-partizan:


last-partizan:
@kishorkunal-raj 

Looks like travis-ci.org is shutting down in the near future, with everything moving to travis-ci.com.

ppc64le architecture is already not working there for some reason
https://travis-ci.org/github/deschler/django-modeltranslation/jobs/767373210

And i'm looking at migration to github actions.

Unfortunately, github actions does not have support for that architecture.

I will not merge this yet, probably will wait to see how travis-ci migrations is going and whather they fix ppc64le runners or not.

PR #595, opened 2021-04-29
Spotted some good old print statements while reading the doc.

JulienPalard:
Thanks for django-modelstranslation while I'm here, I'm using it successfully on hackinscience.org since september 2020 according to git log ☺


last-partizan:
Thanks :)

Issue #602, opened 2021-06-22
Tagging main select with a css class


hovi:
Main language `select` switch created here:

https://github.com/deschler/django-modeltranslation/blob/master/modeltranslation/static/modeltranslation/js/tabbed_translation_fields.js#L443

Is plain `select` element with no class or wrapper. It is hard to select (in terms of css selectors) it using css. Would be nice to add specific class or a another mechanism for styling purposes.

last-partizan:
Please, make a PR with this change.

let's call it `id="modeltranslation-main-switch"`

last-partizan:
Released new version with your patch

PR #604, opened 2021-06-27
added css class to main language switch

hovi:
Simple request to add css class to main language switch. I suggest using class rather than ID even if you don't expect more of same elements to be present in the page right now or in near future.
See #602

### DIFF
--- a/modeltranslation/static/modeltranslation/js/tabbed_translation_fields.js
+++ b/modeltranslation/static/modeltranslation/js/tabbed_translation_fields.js
@@ -440,7 +440,7 @@ var google, django, gettext;
 
     var MainSwitch = {
       languages: [],
-      $select: $("<select>"),
+      $select: $("<select class='modeltranslation-main-switch'>"),
 
       init: function (groupedTranslations, tabs) {
         var self = this;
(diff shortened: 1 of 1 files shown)


modeltranslation/static/modeltranslation/js/tabbed_translation_fields.js
```diff
@@ -440,7 +440,7 @@ var google, django, gettext;
 
     var MainSwitch = {
       languages: [],
-      $select: $("<select>"),
+      $select: $("<select class='modeltranslation-main-switch'>"),
```
last-partizan, line 443:
```suggestion
      $select: $("<select id='modeltranslation-main-switch' class='modeltranslation-switch'>"),
```

@hovi why do you think class is better?

It's single main switch. Maybe we should add id and class? :)

hovi: I understand, that from your point of view it is something unique, that is currently only once on the page. 
From css code point of view it gives no benefit to use idinstead of id.
I am happy either way, having both class and id is good compromise. I will modify PR. Should ID and class have same value?

last-partizan: I think they should be different, like in my suggestion. You can just click "Commit suggestion".

Issue #606, opened 2021-07-17
Using TranslationAdmin with different admin template.


hovi:
Happy user of this package. I recently tried switching admin to other templates (currently experimenting with django-jazzmin).
Sadly tabs are not working there. Been debugging for a while and I think it is mostly (or only) because html structure does not match what is expected in JS code.

For example main select expects `h1` to exist in `#content` to append it there (. In case of this template when I removed requirement for inner `h1`, main select started working, but not the rest.
https://github.com/deschler/django-modeltranslation/blob/master/modeltranslation/static/modeltranslation/js/tabbed_translation_fields.js#L466

Now after I already spent some time on this I'd like to know, if there's any documentation on this - what kind of structure is required for javascript to work or any hints, where should I look in javascript?

Think ideal scenario is if those hook elements in the templates could be somehow configurable using css selectors, but I expect that could require some more extensive refactoring. I am happy to create PR with some guidance if we find some easy solution, that can work nicely for everyone.

last-partizan:
Hi

There is no documentation on required structure, it just expects default django strcuture.

You are looking in right direction. If you want to make it work, you need to identify selectors which are different from django defaults, and create something like mapping.


```
const DefaultDjangoSelectors = {
  mainHeader: () => $("#content").find("h1");
};
const Djazzmin = {
  mainHeader: ...
};

const selectors = detectCurrentTemplate(); // this should return appropriate selectors.
```

And then refactor code to use those custom selectors.

hovi:
Alright. If I create these configurations and function to detect template, are you willing to merge it?

One more step further, this configuration can also contain classes that are used to mark created html. Mostly meaning language switching tabs. I might as well use css framework I am using (at least I hope I could do that by just modifying css, one more step would be possibility to change html too, but I don't wanna go that far).

last-partizan:
Yes, i'll merge it.

PR #607, opened 2021-07-18
semi-configurable selection of elements to generate tabs in admin

hovi:
#606 
Draft PR suggesting direction of configurable selectors.

### DIFF
--- a/modeltranslation/static/modeltranslation/js/tabbed_translation_fields.js
+++ b/modeltranslation/static/modeltranslation/js/tabbed_translation_fields.js
@@ -11,6 +11,34 @@ var google, django, gettext;
   };
 
   jQuery(function ($) {
+    var detectTemplate = function() {
+      if (document.querySelector("#jazzmin-theme, #jazzy-navbar, #jazzy-tabs, #jazzy-actions")) {
+        return "jazzmin";
+      }
+      return "default";
+    }
+
+    var selectorMapping = {
+      "default": {
+        "mainHeader": () => $("#content").find("h1"),
+        "tabContainer": (el) => $(el).closest(".form-row"),
+        "tabUlClass": "",
+        "tabLiClass": "",
+        "tabAClass": "",
+        "tabErrorClass": "ui-tab-has-errors"
+      },
+      "jazzmin": {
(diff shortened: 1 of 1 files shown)


modeltranslation/static/modeltranslation/js/tabbed_translation_fields.js
```diff
@@ -212,9 +240,9 @@ var google, django, gettext;
           container.find("script").remove();
           panel = $('<div id="' + tabId + '"></div>').append(container);
           tab = $(
-            "<li" +
-              (label.hasClass("required") ? ' class="required"' : "") +
-              '><a href="#' +
+            `<li class='${selectors["tabLiClass"]} ` +
+              (label.hasClass("required") ? "required" : "") + "'" +
+              `><a class="${selectors["tabAClass"]}" href="#` +
```
last-partizan, line 245: I think it looks good, except that now we have string templates mixed with concatenation.

But that's for another refactoring.

last-partizan (commented):
I think i can merge it in this state.

Does it work fine for you with both, django and jazzmin templates?
Do you have any other improvements in mind?

hovi:
I wouldn't merge it as it is. It's just a random draft.
It works for me, but it's not very flexible for general purposes.

In jazzmin, main select is visible, but doesn't change languages for tabs, I have to fix that..

I will think about it a bit more. I guess we could merge and then go for better separation/configuration incrementally in different PRs.

hovi:
I think this is good enough for incremental merge. Go ahead and merge if you are okay with it.

A few notes:

- js tab library also puts it's own classes and there's no way to configure "active" tab class.
- There's problem with main select in jazzmin due to using `select2` widget. Not sure exactly why, could also be because of multiple jquery instances.
- Overall I'd say correct solution would be big refactoring with clear separation of functionality and templating. Ideal would be to make js library plugable so that js tab can be used or something else too (ex. bootstrap javascript) or make own implementation, since it should be something very very simple - just hiding/showing content based on active tab.

last-partizan:
Merged it, thanks.

I think to refactor this is a good idea, nobody has touched it for a long time.

PR #618, opened 2021-11-08
Avoid deprecation warning in Django 3.2+

jpmckinney:
I followed the same approach as in https://github.com/jazzband/django-debug-toolbar/pull/1467/files

### DIFF
--- a/modeltranslation/__init__.py
+++ b/modeltranslation/__init__.py
@@ -3,8 +3,12 @@
 Version code adopted from Django development version.
 https://github.com/django/django
 """
+import django
+
 VERSION = (0, 17, 3, 'final', 0)
-default_app_config = 'modeltranslation.apps.ModeltranslationConfig'
+
+if django.VERSION < (3, 2):
+    default_app_config = 'modeltranslation.apps.ModeltranslationConfig'
 
 
 def get_version(version=None):


last-partizan:
Thanks, merged.

last-partizan:
I reverted this, becouse it was breaking install.

But i think it isn't really necessary, besides removing warning. It should work just fine with 4.0.

BSVogler:
Hey last-partizan. How was it breaking the install? Do you have more information to share?

This is fine for now but on the release of the next minor version of Django it will block an update. 
(I would also welcome to have one less error message in the unittest.)

last-partizan:
Here is related issue:

https://github.com/deschler/django-modeltranslation/issues/624

Are you catching deprecation warnings in your tests? I think this particular one could be ignored, becouse it will just work in django-4.

BSVogler:
I think the actual issue is that the setup.py is importing [the init](https://github.com/deschler/django-modeltranslation/blob/master/modeltranslation/__init__.py) to get the version number. This means that no import statement is allowed in the ini file.

I am not sure if deriving the PEP 386-compliant version number in there is some pattern but because of the side effect, it feels like an anti-pattern to me.

jpmckinney:
Indeed, imports from setup.py often cause issues. Just update the version in more than one file 🤷🏼‍♂️. Or use https://pypi.org/project/bumpversion/ if you don’t like that idea.

PR #633, opened 2022-04-29
Preserve annotate() fields in queryset

devxoul:
`MultilingualQuerysetManager` doesn't respect dynamically selected fields using `annotate()`. This PR adds an availability to preserve annotated fields.

### DIFF
--- a/modeltranslation/manager.py
+++ b/modeltranslation/manager.py
@@ -415,10 +415,12 @@ class MultilingualQuerySet(models.query.QuerySet):
         if not kwargs.pop('prepare', False):
             return super(MultilingualQuerySet, self)._values(*original, **kwargs)
         new_fields, translation_fields = append_fallback(self.model, original)
+        annotation_keys = set(self.query.annotation_select.keys())
+        new_fields.update(annotation_keys)
         clone = super(MultilingualQuerySet, self)._values(*list(new_fields), **kwargs)
         clone.original_fields = tuple(original)
         clone.translation_fields = translation_fields
-        clone.fields_to_del = new_fields - set(original)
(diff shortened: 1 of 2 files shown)


devxoul:
@last-partizan Thanks for the quick review! I fixed the lint error :)

rastytheamateur:
Hi,

by this, you have broken `values_list()` method.

In previous version `0.17.5`, it was working well:
```
MyModel.objects.filter(relatedmodel__isnull=False).annotate(Count('id')).values_list(
    'id', 'name').order_by('name').query
>>> SELECT "app_mymodel"."name_en", "app_mymodel"."id" FROM "app_mymodel"
    INNER JOIN "app_relatedmodel" ON ("app_mymodel"."id" = "app_relatedmodel"."mymodel_id")
    WHERE "app_relatedmodel"."id" IS NOT NULL
    GROUP BY "app_mymodel"."id"
    ORDER BY "app_mymodel"."name_en" ASC
```
while in the newest version `0.17.6` you receive by the very same query a different SQL and results as well:
```
MyModel.objects.filter(relatedmodel__isnull=False).annotate(Count('id')).values_list(
    'id', 'name').order_by('name').query
>>> SELECT "app_mymodel"."id", "app_mymodel"."name_en", COUNT("app_mymodel"."id") AS "id__count" FROM "app_mymodel"
    INNER JOIN "app_relatedmodel" ON ("app_mymodel"."id" = "app_relatedmodel"."mymodel_id")
    WHERE "app_relatedmodel"."id" IS NOT NULL
    GROUP BY "app_mymodel"."id"
    ORDER BY "app_mymodel"."name_en" ASC
```
Note the difference in the `SELECT` clause. Incorrect order, incorrect number of fields and one incorrect field at all.

PR #634, opened 2022-04-29
Do not include annotation fields when selects specific fields

devxoul:
After #633 got merged, I found a non-standard behavior. By default, with using `values()` or `values_list()` with specific fields, the queryset should not include annotated fields.

* `queryset.annotate(foo=bar).values()`: `{'id': 10, 'foo': 'bar'}`  # should include annotated fields
* `queryset.annotate(foo=bar).values('id')`: `{'id': 10}`  # should _**NOT**_ include annotated fields

This PR ensures the Django's default behavior of `values()` and `values_list()` with fields.

### DIFF
--- a/modeltranslation/manager.py
+++ b/modeltranslation/manager.py
@@ -414,8 +414,9 @@ class MultilingualQuerySet(models.query.QuerySet):
     def _values(self, *original, **kwargs):
         if not kwargs.pop('prepare', False):
             return super(MultilingualQuerySet, self)._values(*original, **kwargs)
+        selects_all = kwargs.pop('selects_all', False)
         new_fields, translation_fields = append_fallback(self.model, original)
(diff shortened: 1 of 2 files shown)


modeltranslation/manager.py
```diff
@@ -414,8 +414,9 @@ def raw_values(self, *fields):
     def _values(self, *original, **kwargs):
         if not kwargs.pop('prepare', False):
             return super(MultilingualQuerySet, self)._values(*original, **kwargs)
+        selects_all = kwargs.pop('selects_all', False)
```
last-partizan, line 417: i think `selects_all` should be defined explicitly in function definition, becouse otherwise it could be handed over to super()

last-partizan: or, at least, this line should be above 415, if we want to match original definition.

PR #653, opened 2022-09-26
Add rest framework integration. Requires lang in query URL param

chvmq:


contmp:
You should send desired language in header when using drf. Model Translation will already handle throw translated value when using ModelSerializer

last-partizan:
Yes, this isn't required at all. Use `Accept-Language` header.

https://docs.djangoproject.com/en/4.1/topics/i18n/

PR #654, opened 2022-10-12
Support UserAdmin add_fieldsets

napmn:
This PR adds support for `add_fieldsets` when using UserAdmin. Current behaviour is that this attribute is ignored even though it's supported by Django. (https://docs.djangoproject.com/en/4.1/topics/auth/customizing/#custom-users-and-django-contrib-admin).

### DIFF
--- a/modeltranslation/admin.py
+++ b/modeltranslation/admin.py
@@ -34,8 +34,13 @@ class TranslationBaseModelAdmin(BaseModelAdmin):
         # Take custom modelform fields option into account
         if not self.fields and hasattr(self.form, '_meta') and self.form._meta.fields:
             self.fields = self.form._meta.fields
-        if self.fieldsets:
-            return self._patch_fieldsets(self.fieldsets)
+        fieldsets = (
+            self.fieldsets
+            if getattr(self, 'add_fieldsets', None) is None or obj
+            else self.add_fieldsets
+        )
+        if fieldsets:
+            return self._patch_fieldsets(fieldsets)
         elif self.fields:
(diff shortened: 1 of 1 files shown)


modeltranslation/admin.py
```diff
@@ -34,8 +34,13 @@ def _get_declared_fieldsets(self, request, obj=None):
         # Take custom modelform fields option into account
         if not self.fields and hasattr(self.form, '_meta') and self.form._meta.fields:
             self.fields = self.form._meta.fields
-        if self.fieldsets:
-            return self._patch_fieldsets(self.fieldsets)
+        fieldsets = (
```
last-partizan, line 37: Maybe we can use get_fieldsets here?

It would be better, because `add_fieldsets` is not a standard ModelAdmin attribute. UserAdmin overrides get_fieldsets to use this attribute when creating a user. (source: https://docs.djangoproject.com/en/4.1/topics/auth/customizing/)

napmn: We can't use `get_fieldsets` here because `_get_declared_fieldsets` (where these changes are made) is called by `get_fieldsets` in `TranslationAdmin` class. I suppose we could move it to `get_fieldsets` and pass it as argument to `_get_fieldsets_pre_form_or_formset` and `_get_declared_fieldsets`?.

last-partizan: Hmm, no, let's use your original code then. I think it's fine.

But, please add a comment about `add_fieldsets` being non-standard attribute.

napmn: sure, will do

Issue #438, opened 2017-11-27
Thread / eventlet safety of context managers?


rmmr:
I was looking at the inner workings of the translation fallback mechanisms, mostly looking at the helper functions defined at /utils.py . I was curious how you handled concurrency for the context manager `fallbacks()`. This manager function directly modifies settings.ENABLE_FALLBACKS without accounting for any concurrency?

This won't cause any problems for sync workers, but both threaded and async workers can (and most likely will) run into problems. I highly recommend the use of of a Thread Local's object. If you look at Django's source code you will find plently of examples e.g. https://github.com/django/django/blob/5a6f70b4281817656db2f36c5919036d38fcce7f/django/core/cache/__init__.py

zlorf:
Good point.

PR #657, opened 2022-11-05
Make Context Managers (`fallbacks` and `auto_populate`) Thread Safe

saadmk11:
We have been facing issues with `django-modeltranslation` while running **Django** with **Gunicorn** in **multi-threaded** mode.

We were using `fallbacks` context manager and we would sometimes see that all the translation fields in some pages shows `None` even though `MODELTRANSLATION_ENABLE_FALLBACKS` was set to `True`. After some investigation we found that `fallbacks` context manager is changing `MODELTRANSLATION_ENABLE_FALLBACKS` global settings dynamically which meant that another thread accessing the same settings at the same time will receive the changed value.

After looking around the issue tracker we found a related issue and this PR tries to solve it.

closes https://github.com/deschler/django-modeltranslation/issues/438

### DIFF
--- a/modeltranslation/thread_context.py
+++ b/modeltranslation/thread_context.py
@@ -1,28 +1,33 @@
 import threading
+from typing import Literal, Union
 
 from modeltranslation import settings
 
 
+AutoPopulate = Union[bool, Literal["all", "default", "required"]]
+
+
 class ModelTranslationThreadLocal(threading.local):
     """Holds thread-local data for modeltranslation."""
-    auto_populate = None
-    enable_fallbacks = None
+
+    auto_populate: Union[AutoPopulate, None] = None
+    enable_fallbacks: Union[bool, None] = None
 
 
 _mt_thread_context = ModelTranslationThreadLocal()
 
 
-def set_auto_populate(value):
+def set_auto_populate(value: Union[AutoPopulate, None]) -> None:
     """Set the auto_populate for the current thread."""
     _mt_thread_context.auto_populate = value
 
 
-def set_enable_fallbacks(value):
+def set_enable_fallbacks(value: Union[bool, None]) -> None:
     """Set the enable_fallbacks for the current thread."""
     _mt_thread_context.enable_fallbacks = value
 
 
-def auto_populate_mode():
+def auto_populate_mode() -> AutoPopulate:
     """Return the auto_populate mode for the current thread."""
     auto_populate = _mt_thread_context.auto_populate
 
@@ -32,7 +37,7 @@ def auto_populate_mode():
     return settings.AUTO_POPULATE
 
 
-def fallbacks_enabled():
+def fallbacks_enabled() -> bool:
     """Return whether fallbacks are enabled for the current thread."""
     enable_fallbacks = _mt_thread_context.enable_fallbacks
 
--- a/modeltranslation/thread_context.py
+++ b/modeltranslation/thread_context.py
@@ -1,0 +1,42 @@
+import threading
+
+from modeltranslation import settings
+
+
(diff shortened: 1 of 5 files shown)


modeltranslation/thread_context.py
```diff
@@ -0,0 +1,42 @@
+import threading
+
+from modeltranslation import settings
+
+
+class ModelTranslationThreadLocal(threading.local):
+    """Holds thread-local data for modeltranslation."""
+    auto_populate = None
```
last-partizan, line 8: maybe it's better to set default values from settings?

```suggestion
    auto_populate: Union[bool, Literal['all', 'default', 'required']] = settings.AUTO_POPULATE
```

so we can remove "is not None" checks from getters.

And let's try to provide types for new code.

modeltranslation/utils.py
```diff
@@ -141,12 +147,11 @@ def auto_populate(mode='all'):
...
             call_command('loaddata', 'fixture.json')
     """
-    current_population_mode = settings.AUTO_POPULATE
-    settings.AUTO_POPULATE = mode
+    set_auto_populate(mode)
     try:
         yield
     finally:
-        settings.AUTO_POPULATE = current_population_mode
+        set_auto_populate(None)
```
last-partizan, line 154: It isn't "None", it's supposed to revert to previous value.

modeltranslation/utils.py
```diff
@@ -163,12 +168,11 @@ def fallbacks(enable=True):
...
     knowing the language)
     """
-    current_enable_fallbacks = settings.ENABLE_FALLBACKS
-    settings.ENABLE_FALLBACKS = enable
+    set_enable_fallbacks(enable)
     try:
         yield
     finally:
-        settings.ENABLE_FALLBACKS = current_enable_fallbacks
+        set_enable_fallbacks(None)
```
last-partizan, line 175:
```suggestion
        set_enable_fallbacks(None)
```

same here, need's to be reverted to previous.

last-partizan (changes requested):

modeltranslation/thread_context.py
saadmk11: @last-partizan Not setting default value was intensional here. Setting default value will make `fallbacks_enabled()` and `auto_populate_mode()` always return the value from thread local (unless context manager is used). And this will break in 2 places:

1. In tests `seetings.override` will not work as expected for `ENABLE_FALLBACKS` and `AUTO_POPULATE`.
2. if for some reason user changes theses settings at run time the thread locals value will not change. So the settings change will not be effective.

Please let me know if this makes sense. :)

modeltranslation/utils.py
saadmk11: This will remove the value from thread local which will make fallbacks_enabled() and auto_populate_mode() return the value from settings

modeltranslation/thread_context.py
saadmk11: Similar to:
https://github.com/django/django/blob/967f8750ab63f9ca74ce63ada580ccc5b10b3f3b/django/utils/translation/trans_real.py#L320

https://github.com/django/django/blob/967f8750ab63f9ca74ce63ada580ccc5b10b3f3b/django/utils/translation/trans_real.py#L301

modeltranslation/thread_context.py
```diff
@@ -0,0 +1,47 @@
+import threading
+from typing import Literal, Union
```
last-partizan, line 2:
```suggestion
from typing import Union
from typing_extensions import Literal
```

last-partizan:
Merging this as is, and will add `typing_extensions` dependency myself.

PR #668, opened 2023-01-23
Added support to ManyToManyFields 🧑‍🤝‍🧑

bnznamco:
I've added support to translate M2M fields.
I know that supporting relational fields can be a little controversial.
This patch is a step forward in the path that has brought the support to FK and OneToOne fields. 

This patch supports:
 - simple field with explicit or implicit related_name
 - multiple symmetrical field (aka "self") on same model
 - explicit "through model" with custom methods and properties
 - explicit "through model" with translated fields (registered with translator)
 - auto-popolate management command
 
This patch does not support:
 - fallbacks (this is not a technical problem but a conceptual one, do we really need fallbacks on m2m ? What is the definition of empty relation ? Since the field descriptor returns a manager, this could lead to involuntary setting relations on wrong manager. I'm open to suggestions)

I hope you appreciate the work I've done and I'm confident it will be merged soon.
Feel free to ask any question about the implementation, I'm open to review the code with you.

Bye, :)

### DIFF
--- a/modeltranslation/fields.py
+++ b/modeltranslation/fields.py
@@ -1,19 +1,20 @@
-from django import VERSION
-from django import forms
+import copy
+from typing import Iterable
+
+from django import VERSION, forms
 from django.core.exceptions import ImproperlyConfigured
 from django.db.models import fields
 
 from modeltranslation import settings as mt_settings
 from modeltranslation.thread_context import fallbacks_enabled
 from modeltranslation.utils import (
-    get_language,
     build_localized_fieldname,
+    build_localized_intermediary_model,
     build_localized_verbose_name,
+    get_language,
     resolution_order,
-    patch_intermediary_model,
 )
 from modeltranslation.widgets import ClearableWidgetWrapper
-
 
 SUPPORTED_FIELDS = (
     fields.CharField,
@@ -85,7 +86,7 @@ def field_factory(baseclass):
     return TranslationFieldSpecific
 
 
-class TranslationField(object):
+class TranslationField:
     """
     The translation field functions as a proxy to the original field which is
     wrapped.
@@ -162,7 +163,6 @@ class TranslationField(object):
         if isinstance(self.translated_field, fields.related.ManyToManyField) and hasattr(
             self.remote_field, "through"
         ):
-            import copy
 
             # Since fields cannot share the same remote_field object:
             self.remote_field = copy.copy(self.remote_field)
@@ -197,15 +197,16 @@ class TranslationField(object):
                     )
 
             # Patch intermediary model with language scope to create correct db table
-            self.remote_field.through = patch_intermediary_model(self, language, translator)
-            self.remote_field.field = self  # Django 1.6
+            self.remote_field.through = build_localized_intermediary_model(
+                self.remote_field.through, language
+            )
+            self.remote_field.field = self
 
             if hasattr(self.remote_field.model._meta, '_related_objects_cache'):
                 del self.remote_field.model._meta._related_objects_cache
 
         # ForeignKey support - rewrite related_name
         elif not NEW_RELATED_API and self.rel and self.related and not self.rel.is_hidden():
-            import copy
 
             current = self.related.get_accessor_name()
             self.rel = copy.copy(self.rel)  # Since fields cannot share the same rel object.
@@ -219,11 +220,10 @@ class TranslationField(object):
                 )
                 self.related_query_name = lambda: loc_related_query_name
             self.rel.related_name = build_localized_fieldname(current, self.language)
-            self.rel.field = self  # Django 1.6
+            self.rel.field = self
             if hasattr(self.rel.to._meta, '_related_objects_cache'):
                 del self.rel.to._meta._related_objects_cache
         elif NEW_RELATED_API and self.remote_field and not self.remote_field.is_hidden():
-            import copy
 
             current = self.remote_field.get_accessor_name()
             # Since fields cannot share the same rel object:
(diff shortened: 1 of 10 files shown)


modeltranslation/fields.py
```diff
@@ -156,8 +158,53 @@ def __init__(self, translated_field, language, empty_value, *args, **kwargs):
...
+                        build_localized_fieldname(self.model.__name__.lower(), language),
+                    )
+                else:
+                    self.remote_field.related_name = build_localized_fieldname(
+                        self.remote_field.get_accessor_name(), language
+                    )
+
+            # Patch intermediary model with language scope to create correct db table
+            self.remote_field.through = patch_intermediary_model(self, language, translator)
+            self.remote_field.field = self  # Django 1.6
```
last-partizan, line 201: Django 1.6 is not supported, minimun Django version is 3.2

modeltranslation/fields.py
```diff
@@ -156,8 +158,53 @@ def __init__(self, translated_field, language, empty_value, *args, **kwargs):
         # (will show up e.g. in the admin).
         self.verbose_name = build_localized_verbose_name(translated_field.verbose_name, language)
 
+        # M2M support - <rewrite related_name> <patch intermediary model>
+        if isinstance(self.translated_field, fields.related.ManyToManyField) and hasattr(
+            self.remote_field, "through"
+        ):
+            import copy
```
last-partizan, line 165: Please, move this import to the top level

modeltranslation/fields.py
```diff
@@ -156,8 +158,53 @@ def __init__(self, translated_field, language, empty_value, *args, **kwargs):
...
+            self.remote_field.through = patch_intermediary_model(self, language, translator)
+            self.remote_field.field = self  # Django 1.6
+
+            if hasattr(self.remote_field.model._meta, '_related_objects_cache'):
+                del self.remote_field.model._meta._related_objects_cache
+
         # ForeignKey support - rewrite related_name
-        if not NEW_RELATED_API and self.rel and self.related and not self.rel.is_hidden():
+        elif not NEW_RELATED_API and self.rel and self.related and not self.rel.is_hidden():
             import copy
```
last-partizan, line 208: Oh, probably that's why you imported `copy` locally. Let's move this to top-level also.

modeltranslation/fields.py
```diff
@@ -400,6 +447,27 @@ def __get__(self, instance, owner):
...
+    """
+    A descriptor used to return correct related manager without language fallbacks.
+    """
+
+    def __init__(self, field_name, fallback_languages):
+        self.field_name = field_name  # The name of the original field
+        self.fallback_languages = fallback_languages
+
+    def __get__(self, instance, owner):
+        # TODO: do we really need to handle fallbacks with m2m relations?
```
last-partizan, line 460: I'm fine with leaving them without fallbacks, but we must document reasoning behind this choice.

modeltranslation/fields.py
```diff
@@ -400,6 +447,27 @@ def __get__(self, instance, owner):
         return None
 
 
+class TranslatedManyToManyDescriptor(object):
```
last-partizan, line 450:
```suggestion
class TranslatedManyToManyDescriptor:
```

We don't need this `object` in python3.

modeltranslation/fields.py
```diff
@@ -400,6 +447,27 @@ def __get__(self, instance, owner):
...
+    """
+    A descriptor used to return correct related manager without language fallbacks.
+    """
+
+    def __init__(self, field_name, fallback_languages):
```
last-partizan, line 455:
```suggestion
    def __init__(self, field_name: str, fallback_languages: Iterable[str]):
```

Please, add types to as much new code as possible. It helps understand what's going on.

modeltranslation/utils.py
```diff
@@ -183,3 +182,46 @@ def parse_field(setting, field_name, default):
         return setting.get(field_name, default)
     else:
         return setting
+
+
+def patch_intermediary_model(field, lang, translator):
```
last-partizan, line 187:
```suggestion
def patch_intermediary_model(field: Field, lang: str, translator: TranslatorClassHere):
```

last-partizan (changes requested):
Looks good. Some small changes still required, and we need to fix tests.

Btw, what's your use case for this?

modeltranslation/fields.py
bnznamco: Yep

modeltranslation/fields.py
bnznamco: That's the reason I'll move everything. I tried to mimic as much as possible the code I've found.

modeltranslation/fields.py
bnznamco: Yes that's fair, I'll add types as much as possible.

bnznamco:
> Btw, what's your use case for this?

I want to use this in the cms (yes yet another) I'm developing for my [company](https://github.com/lotrekagency).
We where using hvad, with the multi table approach but it creates to much expensive queries especially on relations.
The use-cases we've found are a little edge-cases but we had to deal with them:  for example if you are building a website with products and related product, it can happen that you want to have different relations in different languages due to market specific reasons. I know that "market specific reasons" not necessarily should be resolved with i18n, but we often use country codes as languages.

 > we need to fix tests.

Yes sorry for that. I tried on my local machine with pytest defaults and all was ok. I also wrote specific tests for the m2m field cases. I'll check on all test matrix on my fork and fix that. 

Let's catch up for updates. Thanks!

modeltranslation/fields.py
last-partizan: Whatever you see fitting. Caveats is fine.

bnznamco:
I've made all changes you requested and fixed tests. The only thing I've to do is the documentation part , can you point me to the right section where to write the m2m fallbacks explanation ?

last-partizan:
> I've made all changes you requested and fixed tests. The only thing I've to do is the documentation part , can you point me to the right section where to write the m2m fallbacks explanation ?

[docs](https://github.com/deschler/django-modeltranslation/tree/master/docs)/[modeltranslation](https://github.com/deschler/django-modeltranslation/tree/master/docs/modeltranslation)/caveats.rst

last-partizan:
Released beta version `v0.18.8-beta.1` with your changes.

bnznamco:
Wow! Great News! Thank you for the support! 🎉

Issue #682, opened 2023-05-23
Django 4.2: update_or_create doesn't update translated fields


andrei-shabanski:
`update_or_create()` doesn't update a translated field if the original field name (without language code) is used in `defaults` argument.

The test below passes in Django 4.1, but not in Django 4.2

```python
    def test_update_or_create(self):
        obj = models.TestModel.objects.create(title_de='old de', title_en='old en')

        obj_1, created = models.TestModel.objects.update_or_create(pk=obj.pk, defaults={'title': 'NEW DE TITLE'})
        assert created is False
        assert obj_1.title == 'NEW DE TITLE'
        assert obj_1.title_en == 'old en'
        assert obj_1.title_de == 'NEW DE TITLE'

        # But if we fetch the real object from db,
        # we will see that `title` hasn't been updated
        obj_2 = models.TestModel.objects.get()
        assert obj_2.title == 'NEW DE TITLE'  # AssertionError: assert 'old de' == 'NEW DE TITLE'
        assert obj_2.title_en == 'old en'
        assert obj_2.title_de == 'NEW DE TITLE'  # AssertionError: assert 'old de' == 'NEW DE TITLE'
```

last-partizan:
Hello, can you try to fix this issue and make a PR?

Hopiu:
@andrei-shabanski, can you confirm if you're currently working on this? If not, I'd be happy to give it a try, provided I can complete it within a reasonable timeframe.

andrei-shabanski:
@Hopiu I'm not working on that yet. Feel free to take a look at the issue

last-partizan:
Please test beta version: `v0.18.10-beta.0`.

andrei-shabanski:
`update_or_create()` is working now.

But generally speaking, should this issue be fixed in the save() method? `obj.save(update_fields=['title'])` still updates only the `title` field, not `title_de`.

last-partizan:
Probably, yes. You can try fixing it and make PR.

Hopiu:
I have only limited knowledge of the codebase. I am hesitant to try fixing the issue in the model's save() method since the responsibility lies with the manager and fields. I'd probably attach a `pre_save` signal to check if `update_fields` is present. However, I prefer to avoid signals if possible. @last-partizan, do you have any other suggestions?

last-partizan:
Take a look at the django `Model.save` method, looks like it's using queryset's `_update()` method with values, so we can try to override that method.

[...]

PR #683, opened 2023-05-30
Add update_or_create method to MultilingualQuerySet ...

Hopiu:
And added a test case in ModeltranslationTest class as mentioned in #682 .

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,8 +40,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python: ["3.8", "3.9", "3.10"]
-        django: ["3.2", "4.0"]
+        python: ["3.8", "3.9", "3.10", "3.11"]
+        django: ["3.2", "4.0", "4.2"]
         database: ["sqlite", "postgres", "mysql"]
         include:
           - python: "3.7"
--- a/modeltranslation/manager.py
+++ b/modeltranslation/manager.py
@@ -243,6 +243,21 @@ class MultilingualQuerySet(QuerySet):
                 new_args.append(rewrite_lookup_key(self.model, key))
         return super(MultilingualQuerySet, self).select_related(*new_args, **kwargs)
 
+    def update_or_create(self, defaults=None, **kwargs):
+        """
+        Updates or creates a database record with the specified kwargs. The method first
+        rewrites the keys in the defaults dictionary using a custom function named
+        `rewrite_lookup_key`. This ensures that the keys are valid for the current model
+        before calling the inherited update_or_create() method from the super class.
+        Returns the updated or created model instance.
+        """
+        if defaults is not None:
+            rewritten_defaults = {}
+            for key, value in defaults.items():
+                rewritten_defaults[rewrite_lookup_key(self.model, key)] = value
+            defaults = rewritten_defaults
+        return super(MultilingualQuerySet, self).update_or_create(defaults=defaults, **kwargs)
+
     # This method was not present in django-linguo
     def _rewrite_col(self, col):
         """Django >= 1.7 column name rewriting"""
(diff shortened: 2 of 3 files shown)


modeltranslation/manager.py
```diff
@@ -243,6 +243,21 @@ def select_related(self, *fields, **kwargs):
...
+        `rewrite_lookup_key`. This ensures that the keys are valid for the current model
+        before calling the inherited update_or_create() method from the super class.
+        Returns the updated or created model instance.
+        """
+        if defaults is not None:
+            rewritten_defaults = {}
+            for key, value in defaults.items():
+                rewritten_defaults[rewrite_lookup_key(self.model, key)] = value
+            defaults = rewritten_defaults
+        return super(MultilingualQuerySet, self).update_or_create(defaults=defaults, **kwargs)
```
last-partizan, line 259:
```suggestion
        return super().update_or_create(defaults=defaults, **kwargs)
```

modeltranslation/tests/tests.py
```diff
@@ -463,6 +463,29 @@ def test_constructor(self):
...
         self._test_constructor(keywords)
 
+    def test_update_or_create(self):
+        """
+        Test that update_or_create works as expected
+        """
+        # Create new object
+        obj = models.TestModel.objects.create(title_de='old de', title_en='old en')
+
+        # Update existing object
```
last-partizan, line 473: Maybe add another test case when created is True?

modeltranslation/tests/tests.py
```diff
@@ -463,6 +463,29 @@ def test_constructor(self):
...
+            pk=obj.pk, defaults={'title': 'NEW DE TITLE'}
+        )
+        assert created is False
+        assert obj_1.title == 'NEW DE TITLE'
+        assert obj_1.title_en == 'old en'
+        assert obj_1.title_de == 'NEW DE TITLE'
+
+        # Check that the translation fields are correctly saved and provide the
+        # correct value when retrieving them again.
+        obj_2 = models.TestModel.objects.get()
```
last-partizan, line 484: That's the same objects? Maybe use `instance` variable, and `instance.refresh_from_db()` here?

.github/workflows/test.yml
```diff
@@ -98,4 +98,4 @@ jobs:
           pip install typing-extensions coverage pytest pytest-django pytest-cov $(./get-django-version.py ${{ matrix.django }})
       - name: Run tests
         run: |
-          pytest --cov-report term
+          pytest --cov-report term
```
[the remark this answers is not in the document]
Hopiu: I updated the test.yml and added the missing newline at the end of the file.

modeltranslation/manager.py
Hopiu: Those explicit calls to parent via `super(MultilingualQuerySet, self)` are all over the place, should I start to clean them up?

modeltranslation/tests/tests.py
Hopiu: O.k., I added anotherone.

modeltranslation/tests/tests.py
Hopiu: I'm not sure if I understand what you mean by "instance", the name of the variable? I updated the names and use `refresh_from_db()` now.

modeltranslation/manager.py
last-partizan: If you want to, after merging this you can create new PR.

I'd suggest using `pyupgrade --keep-percent-format --py38-plus modeltranslation/**/*.py`.

modeltranslation/tests/tests.py
last-partizan: That's right.

modeltranslation/manager.py
last-partizan: Nevermind, already done.

PR #687, opened 2023-06-03
Extend update_fields with translation fields in Model.save()

andrei-shabanski:
An alternative approach to resolving #682

Override `QuerySet._update()` to update translation fields if only an original field is passed in `save()` in argument `update_fields`


modeltranslation/manager.py
```diff
@@ -386,6 +371,24 @@ def update(self, **kwargs):
 
     update.alters_data = True
 
+    def _update(self, values):
+        """
+        This method is called in .save() method to update an existing record.
+        Here we force to update translation fields as well if the original
+        field only is passed in `save()` in argument `update_fields`.
+        """
+        field_names_to_update = {field.name for field, _, _ in values}
```
last-partizan, line 380:
```suggestion
        field_names_to_update = {field.name for field, *_ in values}
```

modeltranslation/manager.py
```diff
@@ -386,6 +371,24 @@ def update(self, **kwargs):
 
     update.alters_data = True
 
+    def _update(self, values):
```
last-partizan, line 374: Add types here.

modeltranslation/manager.py
```diff
@@ -386,6 +371,24 @@ def update(self, **kwargs):
...
+        """
+        This method is called in .save() method to update an existing record.
+        Here we force to update translation fields as well if the original
+        field only is passed in `save()` in argument `update_fields`.
+        """
+        field_names_to_update = {field.name for field, _, _ in values}
+
+        extra_values = []
```
last-partizan, line 382:
```suggestion
        translation_fields = []
```

modeltranslation/manager.py
```diff
@@ -386,6 +371,24 @@ def update(self, **kwargs):
...
+        for field, something, value in values:
+            translatable_field_name = rewrite_lookup_key(self.model, field.name)
```
last-partizan, line 384:
```suggestion
            translation_field_name = rewrite_lookup_key(self.model, field.name)
```

modeltranslation/manager.py
```diff
@@ -386,6 +371,24 @@ def update(self, **kwargs):
...
+    def _update(self, values):
+        """
+        This method is called in .save() method to update an existing record.
+        Here we force to update translation fields as well if the original
+        field only is passed in `save()` in argument `update_fields`.
+        """
+        field_names_to_update = {field.name for field, _, _ in values}
+
+        extra_values = []
+        for field, something, value in values:
```
last-partizan, line 383: `something` is bad name, take a look at django source, to see what's really there and name it accordingly.

modeltranslation/tests/tests.py
```diff
@@ -108,6 +117,20 @@ def tearDown(self):
...
+        """
+        with connection.cursor() as cursor :
+            cursor.execute(
+                f"select {', '.join(expected_fields)} from {instance._meta.db_table} where id = {instance.pk}",
+            )
+            row = dictfetchall(cursor)
+            self.assertEqual(
+                row[0],
+                expected_fields,
+            )
```
last-partizan, line 132:
```suggestion
        actual = type(instance).objects.filter(pk=instance.pk).values(*expected_fields.keys())[0]
        assert actual == expected_fields
```

Maybe something like this is better?

And i think it's better to write it as separate function with two args,  `assert_db_record(instance, {"title": "...", ...})`,

last-partizan (changes requested):

modeltranslation/manager.py
andrei-shabanski: Looks like it should be a `Type[Model]`, but it's always None

modeltranslation/tests/tests.py
andrei-shabanski: Your option doesn't extract exact values from the database. Modeltranslation rewrites a query to fetch a value of the default translation field (`"title_de"`) for an original field(`"title"`)

andrei-shabanski: I mean `Model.objects.values("title")` performs `select title_de from table`

modeltranslation/manager.py
last-partizan: translation_values, then?

modeltranslation/tests/tests.py
last-partizan: Oh, yes, Use `.rewrite(False)` like other tests here.

modeltranslation/tests/tests.py
```diff
@@ -16,7 +16,7 @@
 from django.core.files.storage import default_storage
 from django.core.management import call_command
 from django.core.management.base import CommandError
-from django.db import IntegrityError
+from django.db import connection, IntegrityError
```
last-partizan, line 19:
```suggestion
from django.db import IntegrityError
```

andrei-shabanski:
@last-partizan Tests for django < 4.2  fail because we don't add the original field in `update_field`.

Hence the following code updates only `title_de` field in django 4.2, and both fields `title` & `title_de` in django < 4.2

```python
instance, created = models.TestModel.objects.update_or_create(
     pk=obj.pk, defaults={'title_de': 'NEW DE TITLE 2'}
)
```

Should we update the original field in the code below?

last-partizan:
Are you sure? Looking at `update_or_create` which you removed, it doesn't seem to rewrite original key when passing translated key.

But, for the sake of consistency, i think it's fine to update 'title' when passing `title_de` to the function. But, only when 'de' is the default language. Otherwise it would be unexpected behaviour.

last-partizan:
4.0 tests are also failing, as you can see. So your original suggestion may be incorrect.

andrei-shabanski:
The same errors (or behavior) were relevant for the removed implementation of `update_or_create`. We just didn't have tests to catch them

andrei-shabanski:
> But, for the sake of consistency, i think it's fine to update 'title' when passing `title_de` to the function. But, only when 'de' is the default language. Otherwise it would be unexpected behaviour.

The current django-modeltranslation version saves _the last assigned value to the original field_ into "title" column even when a secondary language is active currently. Check the sample below

```python
>>> obj = TestModel(title='default')
... with override('en'):
...     obj.title = 'English variant'
...     obj.save()
... 
... TestModel.objects \
...     .rewrite(False) \
...     .filter(pk=obj.pk) \
...     .values('title', 'title_de', 'title_en') \
...     .first()

{
    "title": "English variant",
    "title_de": "default",
    "title_en": "English variant",
}
```

So "title" might store not only the default translation. 

Should we check the currently active language to determine when the original field (`title`) must be updated? I mean to update "title" even if "de" isn't default language

last-partizan:
> So "title" might store not only the default translation (-- that looks strange to me).

AFAIK, docs are suggesting "not rely on the original field anymore", becouse it can be incosistent.

If we can make it consistent with minimal changes, then yes, let's do it. But if that would be too much rewriting, we should leave this for separate refactoring - if you want to fix it.

Right now we should focus on solving "Extend update_fields with translation variants", without breaking update_or_create. Maybe we even should exclude "title" from assertions, to get tests to pass.

andrei-shabanski:
I excluded "title" from the assertion. Not sure I can find the right approach now due to my limited knowledge of the codebase

modeltranslation/tests/tests.py
```diff
@@ -463,6 +478,84 @@ def test_constructor(self):
...
+
+        obj.title_de = 'de 6'
+        obj.title_en = 'en 6'
+        obj.save(update_fields=['title_de', 'title_en'])
+        assert_db_record(obj, title='de 5', title_de='de 6', title_en='en 6')
+
+        obj.title_de = 'de 7'
+        obj.title_en = 'en 7'
+        obj.save(update_fields=['title', 'title_de', 'title_en'])
+        assert_db_record(obj, title='de 7', title_de='de 7', title_en='en 7')
```
last-partizan, line 513: each of this block shoiuld be separate `pytest.param()` case, and test should contain just one block like.

modeltranslation/tests/tests.py
```diff
@@ -463,6 +478,84 @@ def test_constructor(self):
...
+
+            obj.title_de = 'de 6'
+            obj.title_en = 'en 6'
+            obj.save(update_fields=['title_de', 'title_en'])
+            assert_db_record(obj, title='en 5', title_de='de 6', title_en='en 6')
+
+            obj.title_de = 'de 7'
+            obj.title_en = 'en 7'
+            obj.save(update_fields=['title', 'title_de', 'title_en'])
+            assert_db_record(obj, title='en 7', title_de='de 7', title_en='en 7')
```
last-partizan, line 547: same here.

modeltranslation/tests/tests.py
```diff
@@ -463,6 +478,84 @@ def test_constructor(self):
...
+
+    def test_save_non_original_translation_field(self):
+        obj = models.TestModel.objects.create(title='old')
+
+        obj.title_en = 'en value'
+        obj.save(update_fields=['title'])
+        assert_db_record(obj, title='old', title_de='old', title_en=None)
+
+        obj.save(update_fields=['title_en'])
+        assert_db_record(obj, title='old', title_de='old', title_en='en value')
```
last-partizan, line 557: i think this two can also be move to pytest.param.

last-partizan:
> I excluded "title" from the assertion. Not sure I can find the right approach now due to my limited knowledge of the codebase

@andrei-shabanski please, add a note about this in the test.

modeltranslation/tests/tests.py
andrei-shabanski: pytest.parametrize isn't compatible with unittest.TestCase. We can use either 3rdparty packages such as https://pypi.org/project/parameterized/ or TestCase.subTest()

modeltranslation/tests/tests.py
andrei-shabanski: I left it as is for more readability

last-partizan:
Please reabse it on latest master, CI was broken because latest mariadb was failing to start.

pyproject.toml
```diff
@@ -17,6 +17,7 @@ typing-extensions = "*"
 pdbpp = "*"
 flake8 = "*"
 black = "*"
+parameterized = "*"
```
last-partizan, line 20: It also needs to be addet to `test.yml`.

andrei-shabanski: Added parameterized to test.yml

PR #692, opened 2023-07-04
Fix Literal import in python<3.8

chlammas:
Fix **Literal** import error when the Python version is under 3.8

`ImportError: cannot import name 'Literal'`


last-partizan:
Hi, python 3.7 reached EOL already, so we don't support it.

If you really need it, please use your own github fork.

Issue #698, opened 2023-08-04
Add a feature to translate the model.fields default value based on field language.


bart-maykin:
Currently when the translation fields are created the default value of a model.field will be translated based on the current value.
It would be very handy if this could be changed to the field language of the currently created translation field.

So if you have the following charfield for example:
`description = models.CharField(
    verbose_name=_("description"),
    default=_("Some english text")
)`

And configured your project to support english and dutch with default (and current) language to be english you get the following fields with values:
```
 'description': 'Some english text',
 'description_en': 'Some english text',
 'description_nl': 'Some english text',
```

The wanted outcome would be based on your translation files to something like this for example:
```
 'description': 'Some english text',
 'description_en': 'Some english text',
 'description_nl': 'Wat nederlandse tekst',
```

last-partizan:
Well, that's an interesting feature.

I'm mostly just maintaining this project to ensure it works with latest django versions, but if you're interested, you can try implementing this yourself and create PR, it should be possible to achive you desired result.

sergei-maertens:
Thank you for the quick feedback! We'll come up with a PR

sergei-maertens:
@last-partizan hi! we've implemented this feature, can you give an indication on when you (or another maintainer) might have time to review it? I don't mean to pressure you, just having an ETA is sufficient for us to take it into account for our own planning.

sergei-maertens:
Awesome!

PR #699, opened 2023-08-30
Add missing migrations for tests app/models

sergei-maertens:
Simply checking out master and running makemigrations produces these missing migrations. Run on the lowest supported versions of Python and Django (3.8 and 3.2 respectively).

Separate PR for this to not distract from the main PR, for which a new model field is added to be able to test the behaviour for #698.


last-partizan:
@sergei-maertens i thin, for the test app it's better to just remove initial migration and recreate it.

And don't forget to run `black modeltranslation *.py` on your changes.

sergei-maertens:
> @sergei-maertens i think, for the test app it's better to just remove initial migration and recreate it.
> 
> And don't forget to run `black modeltranslation *.py` on your changes.

Sure, works for me! I'll update the PR

w/r to black - is it only applied on the app itself? I noticed that the tests themselves do not have consistent code formatting, that's why I didn't run it in the first place as it was unclear that black is used. The contributor docs only make mention of flake8.

sergei-maertens:
Rebased and updated with re-generated initial migration.

modeltranslation/tests/migrations/0001_initial.py
```diff
@@ -1,18 +1,17 @@
...
+import django.db.models.deletion
 import modeltranslation.tests.models
 
 
 class Migration(migrations.Migration):
     initial = True
 
     dependencies = [
-        ('auth', '0001_initial'),
+        ('auth', '0013_group_name_de_group_name_en'),
```
last-partizan, line 14:
```suggestion
        ('auth', '0001_initial'),
```

last-partizan:
> w/r to black - is it only applied on the app itself? I noticed that the tests themselves do not have consistent code formatting

It is applied, but for whatever reason we have `skip-string-normalization = true` in pyproject.toml. I'll fix it after we merge your changes.

modeltranslation/tests/migrations/0001_initial.py
sergei-maertens: Updated!

PR #700, opened 2023-08-30
Feature: support language-specific field defaults

### LINKED ISSUE #698: Add a feature to translate the model.fields default value based on field language.

sergei-maertens:
Closes #698, overlaps with #699 w/r to re-generated initial migrations

When django-modeltranslation creates the translation fields (foo_en, foo_de...) and the base field has a callable default set, any language-aware content is now translated with the field's language.

The mechanism is rather simple - `Field.get_default` is overridden to temporarily activate `field.language`.


modeltranslation/tests/models.py
```diff
@@ -1,16 +1,22 @@
...
+    lang = get_language()
+    return f"dynamic: {lang}"
+
+
 class TestModel(models.Model):
     title = models.CharField(gettext_lazy('title'), max_length=255)
     text = models.TextField(blank=True, null=True)
     url = models.URLField(blank=True, null=True)
     email = models.EmailField(blank=True, null=True)
+    dynamic_default = models.CharField(default=get_dynamic_default, max_length=255)
```
last-partizan, line 19: suggestion,open to discussion:

Maybe it's better to test it with translation instead of custom function?

```suggestion
    dynamic_default = models.CharField(default=_("Hello"), max_length=255)
```

That would require adding locales to test app, but will be exactly what we need to check. Or we can use preexisting traslations from django, we're have `django.contrib.auth` app installed in the settings, so we can use any string from `contrib/auth/locale/en/LC_MESSAGES/django.po`

For example, `_("password")`

modeltranslation/fields.py
```diff
@@ -256,6 +258,17 @@ def __ne__(self, other):
...
+    def get_default(self):
+        with override(self.language):
+            default = super().get_default()
+            # we must *force evaluation* at this point, otherwise the lazy translatable
+            # string is returned and will be evaluated only later when the main language
+            # is activated again (because this context block has exited).
+            #
+            # force_str passes protected types as-is, which includes None, int, float,
+            # datetime...
+            return force_str(default, strings_only=True)
```
sergei-maertens, line 270: @last-partizan @bart-maykin took me some serious digging to figure out why the `gettext_lazy("foo")` variant didn't work as expected, but now I finally understand.

During `Model.__init__`, django populates the field defaults using `field.get_default()` if no value was provided (nothing ordinary here yet). The patch with `with override(self.language):` was correctly working and printing the default in the shell also produced the expected translated string. However, when actually running the tests without breakpoints, only the German translation were every reported.

This is because the project language is set to German and the lazy translated string was being returned all the way - effectively the `translation.override` was not having any effect. The lazy string is returned from `get_default`, and only when values are effectively evaluated (e.g. when doing `instance.foo == "bar"` assertions or when it's being written to the database) the lazy string is evaluated, but it's now using `settings.LANGUAGE_CODE` because the context block was exited.

As far as I can tell the only way to prevent this is by forcing the evaluation sooner and still inside the context manager, using the `force_str` utility. This one has a option where it passes through anything that's not like a string (so it would be safe to use this default with non-string based fields or `None` default values as employed by this library!). Locally the rest of the test suite is passing, which confirms this.

I don't think there are any negative drawbacks to this approach, but of course I could be missing something.

modeltranslation/tests/migrations/0001_initial.py
sergei-maertens, line 1: I've regenerated the migrations, so #699 is now not longer needed if this is merged.

modeltranslation/tests/models.py
sergei-maertens: Patch is updated, closing this conversation. I didn't think this approach was possible but I'm very happy that it is, as it leads to an even cleaner API for users of the library!

modeltranslation/fields.py
last-partizan: Yes, that's fine!

sergei-maertens:
Migration fixes applied here too

last-partizan:
Merged migration fixes, now please rebase this branch or merge master into it to resolve conflicts.

last-partizan:
Looks fine, but gitlab-ci is failing for the reasons unrelated to this PR. I'll try running it again later today.

If that doesn't work, i'll test it locally and merge tomorrow.

modeltranslation/fields.py
last-partizan: Looking at this again, and i noticed migrations are using english version for `dynamic_default_de` field.

Is that okay for now? Or maybe we can move this language-dependent evaluation somewhere where it would work for both runtime code and migrations?

sergei-maertens: I believe this is not something this library can fix because (need to confirm this) django effectively disables real translations during makemigrations. That's also why the english string is baked in everywhere  rather than applying `settings.LANGUAGE_CODE` - it's taking the literal from the code.

If you really need dynamic values to be baked in the migration itself, you must use a callable default and not just a lazy value.

If django didn't do this, it would generate new migrations in your site-packages for any library using gettext-lazy in their model field definitions as soon as your project uses a different language than what they were generated with, which is of course unacceptable.

Probably good to mention this in the documentation?

last-partizan: Yes, explaining this in the docs is great idea.

last-partizan: But, before doing that - take a look at `TranslationField.deconstruct`. I think we can handle this problem there.

modeltranslation/fields.py
```diff
@@ -256,6 +258,17 @@ def __ne__(self, other):
...
         return hash((self.creation_counter, self.language))
 
+    def get_default(self):
+        with override(self.language):
+            default = super().get_default()
+            # we must *force evaluation* at this point, otherwise the lazy translatable
+            # string is returned and will be evaluated only later when the main language
+            # is activated again (because this context block has exited).
+            #
+            # force_str passes protected types as-is, which includes None, int, float,
```
last-partizan, line 268: I wonder how `force_str` handles JSONField, or any other non-standard field.

By the looks of it it will try converting it to string.

Maybe it's better to not use `force_str`, and instead use something like

```
from django.utils.functional import Promise
...

if isinstance(default, Promise):
    default = str(default)
```

modeltranslation/fields.py
```diff
@@ -256,6 +258,17 @@ def __ne__(self, other):
     def __hash__(self):
         return hash((self.creation_counter, self.language))
 
+    def get_default(self):
```
last-partizan, line 261: suggestion: With this approach, we need to force evaluation in two places, here and in `deconstruct`.

Maybe we can do this just once in `__init__`?

modeltranslation/fields.py
sergei-maertens: I don't think so, because the [management command](https://github.com/django/django/blob/cafe7266ee69f7e017ddbc0d440084ace559b04b/django/core/management/commands/makemigrations.py#L105) is decorated with `@no_translations` which even deactivates all translations even if `USE_I18N = True` in the settings.

I suppose that in theory you could re-activate and re-deactivate translations in the `deconstruct`, but it seems to diverge pretty strongly from expected django behaviour. I'm not sure how invasive you want this library to be on this aspect, as the functional use case makes a lot of sense to me indeed.

modeltranslation/fields.py
sergei-maertens: Looking at https://github.com/django/django/blob/cafe7266ee69f7e017ddbc0d440084ace559b04b/django/utils/encoding.py#L67 with the value of JSONField being a dict for example, this would indeed result in `str(dict)`. The suggested approach is probably more robust. I'll make sure to add a test for this.

I think that it might be better at this point to create a dedicated test model for these default-value cases to not pollute other models and tests, what do you think?

modeltranslation/fields.py
sergei-maertens: that would break the behaviour of actual callable `default=some_function`, as the `__init__` of the field is only invoked once (similarly to how `models.CharField(...)` itself executes only once) when the translation field instance is created and added to the model.

E.g. if you were to specify `default=timezone.now`, this would effectively be the same thing as setting `default=timezone.now()` which is *very* different behaviour.

modeltranslation/fields.py
sergei-maertens: For the following supported types of fields, I don't see how a translatable string (via `gettext_lazy` or otherwise) would make sense: IntegerField, BooleanField, FloatField, DecimalField, IPAddressField, GenericIPAddressField, DateField, DateTimeField, TimeField, ForeignKey, ManyToManyField. Sure, you could have different values depending on language, but not something that would benefit from the field language being activated during `get_default`.

That leaves the following fields:

* `CharField`, `TextField`, `FileField` and `ImageField` - straight forward text based fields for which the `force_str` approach works reliably.
* `JSONField`, for which it doesn't work, but we could `isinstance` this and apply the force_str recursively, you only need to support looping over items for arrays, looping over key-values for dicts and handle the primitives, for which you can rely on `force_str` as it handles `None`, numbers and datetime stuff (essentially all the types that django's DjangoJSONEncoder supports).

last-partizan: That would complicate things too much. Let's start with what we have already, and fix our approach only if we encounter problems.

modeltranslation/fields.py
last-partizan: Ok, let's ignore this problem with migrations. If people want translated defaults here, they can always add it manyally.

last-partizan:
Is there anything else you want to check or fix before merging?

It looks fine to me, and i plan merge it tomorrow.

sergei-maertens:
No, let's go for it!

Issue #597, opened 2021-05-31
No tabbed UI for inline admin rows.


DeadTarotWind:
Sometimes this event is not firing and a proper tabbed UI is not loaded.

https://github.com/deschler/django-modeltranslation/blob/10ec4ed8694d949815ccf4ada679a1cb72f24675/modeltranslation/static/modeltranslation/js/tabbed_translation_fields.js#L515

last-partizan:
Hmmm.

Can you try to investigate when? and maybe make PR with fix?

TDawid:
I do not know the justification of the second condition ( $(window).on("load") ), but in my case I've just relied on $(document).ready() and so far I do not observe any problems.

Danolz:
it happens when you use FireFox, and also sometimes on Chrome without plugins

last-partizan:
Please try version `0.18.13b1`

PR #711, opened 2023-11-17
Fixed bug in tabbed_translation_fields.js (issue #597)

john-psina:
Fixed an issue in file tabbed_translation_fields.js where the add new inline event would not fire. The problem occurred mainly in Firefox, but sometimes also occurred in Chrome

Isue #597


last-partizan:
Thanks!

PR #716, opened 2024-01-21
Add typing

### LINKED ISSUE #715: Typing support
Any interest in having typing introduced in the library?

Let me know and I'll come with a PR

Viicos:
Fixes #715

I've made use of `typing-extensions`. I know some people doesn't like to add an extra dependency purely for typing, but `asgiref` (a Django dependency) requires it anyway.

I made a first `mypy` run, and it was looking alright, I'll finish adding typing when I get some more time.


modeltranslation/utils.py
```diff
@@ -13,46 +16,48 @@
...
 
-def get_language():
+def get_language() -> str | None:
     """
     Return an active language code that is guaranteed to be in
     settings.LANGUAGES (Django does not seem to guarantee this for us).
     """
     lang = _get_language()
-    if lang is None:  # Django >= 1.8
-        return settings.DEFAULT_LANGUAGE
```
last-partizan, line 24: This has no relation to adding a types, revert it.

modeltranslation/utils.py
```diff
@@ -13,46 +16,48 @@
...
 
 
-def build_localized_fieldname(field_name, lang):
+def build_localized_fieldname(field_name: str, lang: str) -> str:
     if lang == "id":
         # The 2-letter Indonesian language code is problematic with the
         # current naming scheme as Django foreign keys also add "id" suffix.
         lang = "ind"
-    return str("%s_%s" % (field_name, lang.replace("-", "_")))
+    return f"{field_name}_{lang.replace('-', '_')}"
```
last-partizan, line 57: This is also out of scope of the PR.

Upgrading syntax can be handled automatically with pyupgrade, and is not related to types.

last-partizan:
Overall looks good!

We also need to add mypy to CI

modeltranslation/utils.py
Viicos: Ok after this one I will make two separate PRs:
- Remove old compat code
- Use pyupgrade (via ruff) (and maybe enforce it at the CI level)

modeltranslation/utils.py
```diff
@@ -112,7 +117,9 @@ def unique(seq):
     return (x for x in seq if x not in seen and not seen.add(x))
```
Viicos, line 117: mypy complains with:

`modeltranslation/utils.py:119: error: "add" of "set" does not return a value (it only ever returns None)  [func-returns-value]`

this implementation seems to be a hacky way of adding the element in the set while still keeping it a one liner. Maybe a more explicit implementation should be used instead?

Viicos:
`mypy` is not "smart" enough to infer `FALLBACK_LANGUAGES` as being `dict[str, tuple[str, ...]]`:

https://github.com/deschler/django-modeltranslation/blob/2f256fa34920dc6174178b2ccd182404ac080dee/modeltranslation/settings.py#L40-L42

Even by annotating `FALLBACK_LANGUAGES` in the first line. `pyright` seems to handle it fine. I'll try to find a workaround but that might involve some code changes in the settings module.

This then causes issues when `FALLBACK_LANGUAGES` is being used:

https://github.com/deschler/django-modeltranslation/blob/2f256fa34920dc6174178b2ccd182404ac080dee/modeltranslation/utils.py#L126-L127

modeltranslation/utils.py
last-partizan: For now, add `# type: ignore` for this line

And we can change implementation to non-hacky later.

last-partizan:
> mypy is not "smart" enough to infer FALLBACK_LANGUAGES as being dict[str, tuple[str, ...]]:
> Even by annotating FALLBACK_LANGUAGES in the first line. 

I don't see any annotation for it. Maybe you're annotated another variable?

Viicos:
This ended up being a bit bigger than I expected, but I'm pretty surprised with the reasonable amount of type ignores I had to use (around 35, considering a couple of them will be unnecessary with a new `django-stubs` release).

I still had to globally ignore some codes per modules, because it was just counter productive to actually enable them (we need to remember the primary goal is for end users and also when developing the library. I have to say adding type hints really helped to understand what `modeltranslation` was doing in some cases).

modeltranslation/admin.py
```diff
@@ -1,11 +1,18 @@
+from __future__ import annotations
+
 from copy import deepcopy
+from typing import Any, Iterable, Sequence, cast
 
 from django import forms
+from django.db.models import Field as DbField
```
last-partizan, line 7: Why not using Field?

last-partizan:
I was probably thinking that it's still WIP, because of the title. But looks like it's ready?

I'm gonna review this, you need to resolve small merge conflict in lockfile, and we can merge this.

.github/workflows/test.yml
```diff
@@ -29,6 +29,13 @@ jobs:
       - name: Run linters
         run: |
           make lint
+      - name: Install type checking dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install django==4.2.10 django-stubs==4.2.7 mypy==1.8.0
```
last-partizan, line 35: I see you're using django-stubs, but no mypy.ini with configured plugin. Not using plugin probably ok if we aren't using anything requiring plugin.

But let's add `mypy.ini` with at least this

```ini
[mypy]
incremental = True
warn_unused_ignores = True
warn_redundant_casts = True
warn_unused_configs = True
pretty = True
show_error_context = True
```

modeltranslation/admin.py
```diff
@@ -43,12 +51,12 @@ def _get_declared_fieldsets(self, request, obj=None):
             return [(None, {"fields": self.replace_orig_field(self.get_fields(request, obj))})]
         return None
 
-    def formfield_for_dbfield(self, db_field, request, **kwargs):
+    def formfield_for_dbfield(self, db_field: DbField, request: HttpRequest, **kwargs: Any) -> forms.Field:  # type: ignore[return-value]
         field = super().formfield_for_dbfield(db_field, request, **kwargs)
-        self.patch_translation_field(db_field, field, request, **kwargs)
-        return field
+        self.patch_translation_field(db_field, field, request, **kwargs)  # type: ignore[arg-type]
+        return field  # type: ignore[return-value]
```
last-partizan, line 57: This looks suspicious, i was expecting single `ignore[return-value]` here or on the def line.

That's why we need `warn_unused_ignores = True` for.

modeltranslation/admin.py
```diff
@@ -62,12 +70,12 @@ def patch_translation_field(self, db_field, field, request, **kwargs):
...
             pass
         else:
-            orig_formfield = self.formfield_for_dbfield(orig_field, request, **kwargs)
+            orig_formfield = cast(forms.Field, self.formfield_for_dbfield(orig_field, request, **kwargs))
             field.widget = deepcopy(orig_formfield.widget)
             attrs = field.widget.attrs
             # if any widget attrs are defined on the form they should be copied
             try:
-                field.widget = deepcopy(self.form._meta.widgets[orig_field.name])  # this is a class
+                field.widget = deepcopy(self.form._meta.widgets[orig_field.name])  # type: ignore[index]
```
last-partizan, line 78: Try to preserve comment

pyproject.toml
```diff
@@ -12,19 +12,44 @@ packages = [
...
+mypy = "^1.8.0"
 
 [tool.ruff]
 line-length = 100
 target-version = "py38"
 ignore = [
     "E501",   # line length is handled by formatter
 ]
+
+[tool.mypy]
```
last-partizan, line 34: oh, i see where's mypy config. Nice, let's add some more options here. No need for separate file.

pyproject.toml
```diff
@@ -12,19 +12,44 @@ packages = [
 python = ">=3.8,<4"
 django = ">=3.2"
 ruff = "^0.0.292"
+typing-extensions = {version = ">=4.0.1", python = "3.11"}
```
last-partizan, line 15: we also need to add deps to setup.cfg

this file is used mainly in development (if my memory serves me well)

modeltranslation/admin.py
Viicos: Nice catch, redundant ignore removed.

By the way, the next release of `django-stubs` will remove the need for the last type ignore here

.github/workflows/test.yml
Viicos: TIL about `warn_redundant_casts`, really useful

Viicos:
One mypy failure should remain:

```sh
modeltranslation/admin.py: note: In member "_get_declared_fieldsets" of class "TranslationBaseModelAdmin":
modeltranslation/admin.py:51: error: Argument 1 to "replace_orig_field" of "TranslationBaseModelAdmin" has incompatible type
"Sequence[Union[str, Sequence[str]]]"; expected "Iterable[str]"  [arg-type]
                return [(None, {"fields": self.replace_orig_field(self.get_fields(request, obj))})]
                                                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

and the current signature of `replace_orig_field` is:

```python
def replace_orig_field(self, option: Iterable[str]) -> _ListOrTuple[str]: ...
```

last-partizan:
> One mypy failure should remain:
> 
> ```shell
> modeltranslation/admin.py: note: In member "_get_declared_fieldsets" of class "TranslationBaseModelAdmin":
> modeltranslation/admin.py:51: error: Argument 1 to "replace_orig_field" of "TranslationBaseModelAdmin" has incompatible type
> "Sequence[Union[str, Sequence[str]]]"; expected "Iterable[str]"  [arg-type]
>                 return [(None, {"fields": self.replace_orig_field(self.get_fields(request, obj))})]
>                                                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```
> 
> and the current signature of `replace_orig_field` is:
> 
> ```python
> def replace_orig_field(self, option: Iterable[str]) -> _ListOrTuple[str]: ...
> ```
> 
> Not sure what to do with this one, I don't know if `get_fields` can really return `Sequence[Sequence[str]]`?

mypy is right here, take a look at the function dosctring, it really accepts `Sequence[Sequence[str] | str]`

modeltranslation/admin.py
```diff
@@ -167,7 +167,7 @@ def replace_orig_field(self, option: Iterable[str]) -> _ListOrTuple[str]:
             for opt in option:
                 if opt in self.trans_opts.fields:
                     index = option_new.index(opt)
-                    option_new[index : index + 1] = get_translation_fields(opt)
+                    option_new[index : index + 1] = get_translation_fields(opt)  # type: ignore[arg-type]
```
Viicos, line 170: `pyright` is smart enough to narrow `opt` to `str` thanks to the `if opt in self.trans_opts.fields` check, but not `mypy`

Makefile
```diff
@@ -12,3 +12,6 @@ clean:
 lint:
 	ruff check modeltranslation
 	ruff format --check modeltranslation *.py
+
+typecheck:
+    mypy modeltranslation
```
last-partizan, line 17:
```suggestion
	mypy modeltranslation
```

last-partizan:
Now, please fix linter errors and merge latest changes from master into your branch.

last-partizan:
Looks like we need to apply `ruff format`.

pyrightconfig.json
last-partizan, line 1: And please restore this file.

Viicos: Woops, didn't mean to stage this change

Viicos:
I still need to look into test failures

Viicos:
@last-partizan Sorry if this took some time, I was struggling to find some free time working on it. But this is now done I believe

last-partizan:
@Viicos don't worry, we're all unpaid volunteers here, and nobody expects any deadlines :)

Please, merge master into this branch, and resolve conflicts.

Viicos:
Rebased (merging wasn't possible for some reason (says I'm up to date), the branch is in a weird state)

last-partizan:
Thanks, merged!

Viicos:
Thanks!

Issue #727, opened 2024-04-19
Translated foreign key in an abstract base class:  foreign key instances cannot be deleted.


benzkji:
I have a abstract `Link` base class, used for my `Teaser` and `ContentNav` models. The base class has a field `file`, that is then translated in each of the real classes. The file field is a `FileField` from django-filer, so basically a `ForeignKey`.

When I try to delete a `File` instance, I get a `TypeError / Cannot combine queries on two different base models.` from django itself. The `file` field is configured as `on_delete=models.SET_NULL`, and as far as I can tell, the error happens because django wants to update all the needed foreign key fields, but somehow gets two querysets of different models (the two translated above) to combine and update. Why exactly is very hard to tell.

Has this happend before? I'll try to write a test, to rebuild the error - it currently happens in an old project, that was updated several times.

last-partizan:
I'm not sure.

Do you get this error when you disable django-modeltranslation?

benzkji:
I'm not sure, will try.

benzkji:
Indeed, if I disable translation on one of the translated models, deletion is possible. I'll try to write a test, seems like a very edge case.

benzkji:
please see PR, need directions for adding migrations :)

benzkji:
The PR: #729

PR #729, opened 2024-04-20
#727 wip, added tests, without migrations

benzkji:
I've added tests, but don't know how I could add migrations for the new models needed for testing.

### DIFF
--- a/modeltranslation/tests/translation.py
+++ b/modeltranslation/tests/translation.py
@@ -85,6 +85,22 @@ class ForeignKeyModelTranslationOptions(TranslationOptions):
     )
 
 
+@register(models.FirstSetNullForeignKeyModel)
+class FirstSetNullForeignKeyModelTranslationOptions(TranslationOptions):
+    fields = (
+        "test",
+    )
+
+
+@register(models.SecondSetNullForeignKeyModel)
+class SecondSetNullForeignKeyModel(TranslationOptions):
+    fields = (
+        "test",
+    )
+
+
+
+
 @register(models.OneToOneFieldModel)
 class OneToOneFieldModelTranslationOptions(TranslationOptions):
     fields = (
--- a/modeltranslation/tests/models.py
+++ b/modeltranslation/tests/models.py
@@ -95,6 +95,25 @@ class ForeignKeyModel(models.Model):
         on_delete=models.CASCADE,
     )
 
+class SetNullForeignKeyBaseModel(models.Model):
+    test = models.ForeignKey(
+        TestModel,
+        null=True,
+        related_name="test_fks",
+        on_delete=models.CASCADE,
+    )
+
+    class Meta:
+        abstract = True
+
+
+class FirstSetNullForeignKeyModel(SetNullForeignKeyBaseModel):
+    pass
+
+
+class SecondSetNullForeignKeyModel(SetNullForeignKeyBaseModel):
+    pass
+
 
 class OneToOneFieldModel(models.Model):
     title = models.CharField(gettext_lazy("title"), max_length=255)
--- a/modeltranslation/tests/tests.py
+++ b/modeltranslation/tests/tests.py
@@ -1182,6 +1182,10 @@ class ForeignKeyFieldsTest(ModeltranslationTestBase):
         field = models.ForeignKeyModel._meta.get_field("test")
         assert field.attname != build_localized_fieldname(field.name, "id")
 
+    def test_multiple_translated_foreign_keys_delete_foreign_obj(self):
+        test = TestModel.objects.create(name="test")
+        test.delete()
+
 
 class ManyToManyFieldsTest(ModeltranslationTestBase):
     @classmethod


modeltranslation/tests/translation.py
```diff
@@ -85,6 +85,22 @@ class ForeignKeyModelTranslationOptions(TranslationOptions):
...
+    fields = (
+        "test",
+    )
+
+
+@register(models.SecondSetNullForeignKeyModel)
+class SecondSetNullForeignKeyModel(TranslationOptions):
+    fields = (
+        "test",
+    )
```
last-partizan, line 99: I see the problem now.

You should register for translation your base model, if you're using fields from base model.

Please try it. And if it fixes the issue, we can raise error "ImproperlyConfigured()" or something like that.

benzkji: Yep, that would make it work. If my base model is not under my control, it will not. There probably would be a solution, but the Django code is very complex :|. Let me dig a bit more into...

benzkji: Sorry. Misunderstood. It's this section: https://django-modeltranslation.readthedocs.io/en/latest/registration.html#translationoptions-fields-inheritance

So I would register the abstract base model with it's foreign key for translation. And maye the sub models, with additional translated fields, right?

benzkji: Just found I got a third model, inheriting from base class, without having the specific field translated. That would not be possible anymore...but would probably not be a problem, besides having unused fields.

last-partizan: Main point: you register fields for the model they're defined on.

I'm not sure, if you also need to register child models (with empty fields), try both variants and see which works.

benzkji: I've done this many times without problems (registering only the final model, but not the abstract base class). But yeah, I can see why it doesn't work in my edge case, with foreign key.

benzkji:
I can't get it to work properly. See my latest commit, that's how it's supposed to be setup?

modeltranslation/tests/translation.py
```diff
@@ -85,6 +85,26 @@ class ForeignKeyModelTranslationOptions(TranslationOptions):
...
 
 
+@register(models.SetNullForeignKeyBaseModel)
+class SetNullForeignKeyBaseModelTranslationOptions(TranslationOptions):
+    fields = [
+        "test",
+    ]
+
+@register(models.FirstSetNullForeignKeyModel)
+class FirstSetNullForeignKeyModelTranslationOptions(SetNullForeignKeyBaseModelTranslationOptions, TranslationOptions):
```
last-partizan, line 95: No need to inherit previous class:

image

benzkji: But my base class is abstract?

benzkji: Im my experience, it doesn't matter if the abstract or concrete model is registered. The problem is, that django thinks that the two fields of the child models belong to the same model, and then trigger the `TypeError / Cannot combine queries on two different base models` error, during the update query for the "models.SET_NULL" foreign keys. That's what I think, I might be wrong.

last-partizan: Oh, i missed that part about abstract model. But anyway, you should not inherit options classes.

benzkji: But that's what docs do/say? https://django-modeltranslation.readthedocs.io/en/latest/registration.html#translationoptions-fields-inheritance

Abstract base class: Inherit TranslationOptionsBaseClass. Concrete base class: Don't inherit TranslationOptions.

For me, it doesn't matter, and I will happily do what solves the issue (as you said, translate/register on the model where the field is defined), but for now there is no solution, or I did not proceed correctly. If you run tests, you'll see yourself.

benzkji: And: Thank you for your help @last-partizan really appreciate!

modeltranslation/tests/translation.py
benzkji: None works.I've not found a solution, and provided a working (failing) test. So I guess it is a bug?

last-partizan: Yeah, probably a bug.

benzkji:
Any time/input, here? 

For me, it's a use case of modeltranslation - having an abstract base class with a foreign key, and having multiple classes that inherit from it, each using translated fields. As it's in the tests:
https://github.com/deschler/django-modeltranslation/pull/729/files#diff-a11e634034d805d4ce683d541f9b181dfc9e2d56beb6036134368586fba097d6R98

It's really deep in the Django core, I still think I was right here: "Django thinks that the two fields of the child models belong to the same model, and then trigger the TypeError / Cannot combine queries on two different base models error, during the update query for the "models.SET_NULL" foreign keys. That's what I think, I might be wrong."

So it's somewhere in Metaclasse, wherever hidden deep?

last-partizan:
Hello!

Last time we stopped at the step where it fails just as it should fail, e.q. "issue is present and we have correct test-case for reproducing it". Right?

As for what to do next, I don't know.

Have you tried removing `TranslationOptions` here?

```py
class SecondSetNullForeignKeyModelTranslationOptions(SetNullForeignKeyBaseModelTranslationOptions, TranslationOptions):
    fields = [
        "translated22",
    ]
```

If removing it helps, great. If it doesn't help - try digging deeper, and ask questions. I'll try to share my knowledge, but this project is fairly complex and I just doing maintenance here, I'm not the original creator.

PR #730, opened 2024-04-21
showcase for issue with foreign keys defined in "app.ModelName" format in abstract base classes

benzkji:
Somehow the field.model is not yet a class, but still a string, in this particular case. 

Tried a fix (modeltranslation/fields.py), but issue seems deeper, I guess it should really just not be a string?

This PR just shows the problem.

### DIFF
--- a/modeltranslation/fields.py
+++ b/modeltranslation/fields.py
@@ -249,8 +249,9 @@ class TranslationField:
                 self.related_query_name = lambda: loc_related_query_name
             self.remote_field.related_name = build_localized_fieldname(current, self.language)
             self.remote_field.field = self
-            if hasattr(self.remote_field.model._meta, "_related_objects_cache"):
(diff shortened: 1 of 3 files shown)


benzkji:
This issue arose during #727 and #729

last-partizan:
Probably, yes. We expect there a model, not a string.

benzkji:
But in this edge case, it is a string. Nothing special, or do you see something special in my example code? Defining foreign key field's models as strings is supported and sometimes necessary, in Django, so shouldn't it be with modeltranslation?

last-partizan:
That looks like error, or maybe we're accessing it at the wrong time (during __init__).

If you look at any other foreign key, `field.model` is a model class, even if it's string in the definition.

benzkji:
Agreed. Hard to tackle down. It's really only when in an abstract base class.

Issue #719, opened 2024-02-08
Django 5 support?


Mapiarz:
Is Django 5.0 supported, officially or not? Anybody tried it?

last-partizan:
Haven't tried personally, but it should work.

If you interested - you can fork the repo, see if test suite passes on django 5, and make a PR enabling it in the pipeline. (Maybe fixing some bugs along the way)

last-partizan:
I tried it, and it does not work.

philitell:
Please support Django 5 - that would be really great.

Thanks for this great project!!!

koleror:
Why do you say it doesn't work? I personnally use it with django 5.0.3 and it works just fine

philitell:
I tested it with Django 5.0.4 and Django-Restframework 3.15.1 --> package works perfectly

Thanks for this great project!!!

last-partizan:
> Why do you say it doesn't work? I personnally use it with django 5.0.3 and it works just fine

Test suite was not passing. If it works in general, it's great news. But please be aware - you can still catch some bugs, until we add django-5.0 to our test matrix and all checks are passed.

hovi:
What doesn't work for me is django admin frontend (no tabs). This is chrome console output:
```
TypeError: tabsContainer.tabs is not a function
    createTabs http://localhost:8000/static/modeltranslation/js/tabbed_translation_fields.js:259
    each jQuery
    createTabs http://localhost:8000/static/modeltranslation/js/tabbed_translation_fields.js:209
    <anonymous> http://localhost:8000/static/modeltranslation/js/tabbed_translation_fields.js:525
    jQuery 13
```

last-partizan:
Now we're testing against django 5.0, and it works.

ateuber:
Unfortunately it is not working with Django 5.1
```
  File "/usr/local/lib/python3.11/site-packages/modeltranslation/fields.py", line 221, in __init__
    elif self.remote_field and not self.remote_field.is_hidden():
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'ManyToOneRel' object has no attribute 'is_hidden'. Did you mean: 'hidden'?
```

last-partizan:
Tried to fix this in #756, but it's slightly more complicated than that.

You can investigate why tests are failing in `django-5.1` branch and create PR targeting that branch with a fixes. Or i will look into this when i have more time.

last-partizan:
[...]

PR #733, opened 2024-05-16
fix: Remove deprecated test config starting from Django 5.0

wookkl:
Hello,

I found and resolved the issue with testing not functioning in Django 5.0 version, after seeing the [issue comment](https://github.com/deschler/django-modeltranslation/issues/719#issuecomment-2077809395).

The cause was the `"TEST": {"SERIALIZE": False}` option inside the `_get_database_config` in tests/settings.py. 
This option was deprecated since version [4.0](https://docs.djangoproject.com/en/5.0/releases/4.0/#id2) and removed in version [5.0](https://docs.djangoproject.com/en/5.0/releases/5.0/#features-removed-in-5-0).

Removing `"TEST": {"SERIALIZE": False}` leads to an error related to `GroupTranslationOptions` in tests/translation.py.

```shell
self = <django.db.backends.sqlite3.base.SQLiteCursorWrapper object at 0x7f9c9513b880>
query = 'SELECT "auth_group"."id", "auth_group"."name", "auth_group"."name_de", "auth_group"."name_en" FROM "auth_group" ORDER BY "auth_group"."id" ASC', params = ()

    def execute(self, query, params=None):
        if params is None:
            return super().execute(query)
        # Extract names if params is a mapping, i.e. "pyformat" style is used.
        param_names = list(params) if isinstance(params, Mapping) else None
        query = self.convert_query(query, param_names=param_names)
>       return super().execute(query, params)
E       django.db.utils.OperationalError: no such column: auth_group.name_de
```
To resolve this, I added the `TEST_NONSERIALIZED_APPS` option.

### DIFF
--- a/modeltranslation/tests/settings.py
+++ b/modeltranslation/tests/settings.py
@@ -10,9 +10,6 @@ def _get_database_config():
     conf = {
         "ENGINE": "django.db.backends.sqlite3",
         "NAME": ":memory:",
-        "TEST": {
-            "SERIALIZE": False,
-        },
     }
     if db == "mysql":
         conf.update(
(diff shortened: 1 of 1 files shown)


last-partizan:
That's a good start. Can you please add Django 5 to the test matrix next?

wookkl:
Will do!

PR #737, opened 2024-05-23
fix(types): make admin classes generic as their super classes

mschoettle:
`BaseModelAdmin` and `ModelAdmin` are generic:

https://github.com/typeddjango/django-stubs/blob/496190e1b5dc344980d8f6ee81ff5dcecc05ded9/django-stubs/contrib/admin/options.pyi#L88

https://github.com/typeddjango/django-stubs/blob/496190e1b5dc344980d8f6ee81ff5dcecc05ded9/django-stubs/contrib/admin/options.pyi#L145

This PR makes the translation admin classes generic as well.


last-partizan:
Good start, now let's fix CI :)

Viicos:
You might be interested in this thread: https://github.com/lazybird/django-solo/pull/128#discussion_r1451790020

modeltranslation/admin.py
```diff
@@ -25,8 +30,31 @@
...
+    pass
+
+class ModelAdmin(MonkeyPatchMixin, _ModelAdmin):
+    pass
+
+class StackedInline(MonkeyPatchMixin, _StackedInline):
+    pass
+
+class TabularInline(MonkeyPatchMixin, _TabularInline):
+    pass
```
last-partizan, line 54: No, let's not use this variant.

Let's either add `django-stubs-ext` and call patch somewhere in the `__init__.py`, or add a function `monkeypatch()` with something like:

```
if not hasattr(admin.ModelAdmin, "__class_getitem__"):
    # https://github.com/typeddjango/django-stubs/tree/master/django_stubs_ext
    # For generic classes to work at runtime we need to define `__class_getitem__`.
    # We're defining it here, instead of relying on django_stubs_ext, because
    # we don't want every user setting up django_stubs_ext just for this feature.
    # In theory, this can be replaced with `if TYPE_CHECKING` clause for base class,
    # but mypy does not support it at the time of writing (22 Aug 2023).
    models.CharField.__class_getitem__ = classmethod(  # type: ignore
        lambda cls, *args, **kwargs: cls
    )
```

into _typing.py, and call in in __init__.py.

Viicos: I would go with the second solution, this would avoid an extra dependency on `django-stubs-ext` for a quite small typing improvement :).

> but mypy does not support it at the time of writing (22 Aug 2023).

Just wondering, does this hold true as of today?

last-partizan: Oh, i copied this from my other PR, and that's about making the following:

```python
class PhoneNumberField(CharField[str, str] if TYPE_CHECKING else CharField):
    pass
```

https://github.com/stefanfoulis/django-phonenumber-field/pull/567/files

Yes, it still doesn't work.

```
test.py:5: error: Invalid base class  [misc]
    class PhoneNumberField(CharField[str, str] if TYPE_CHECKING else CharField):
```

But this was fine for PhonenumberField, not for our case. Because here end-users need to use `if TYPE_CHECKING` every time they define a class.

Viicos: Ah yes, I think https://github.com/lazybird/django-solo/pull/128#discussion_r1451790020 is supported by mypy (no inlined `if TYPE_CHECKING` block). But the runtime addition of `__class_getitem__` is also fine, both work perfectly.

mschoettle: I added the second solution as suggested since it's required at runtime.

There are two issues with `TranslationInlineModelAdmin` remaining in regards to `get_formsets`. `InlineModelAdmin` is defined with `obj` being `_ParentModelT` (see: https://github.com/typeddjango/django-stubs/blob/master/django-stubs/contrib/admin/options.pyi#L317) whereas `TranslationBaseModelAdmin._get_form_or_formset` and `get_fieldsets` have `obj: _ChildModelT | None`.

Does either of you have any ideas on how this could be resolved?

modeltranslation/_typing.py
```diff
@@ -14,3 +17,22 @@
...
 # See https://github.com/typeddjango/django-stubs/blob/082955/django-stubs/utils/datastructures.pyi#L12-L14
 _ListOrTuple: TypeAlias = "list[_K] | tuple[_K, ...]"
+
+
+# https://github.com/typeddjango/django-stubs/tree/master/django_stubs_ext
+# For generic classes to work at runtime we need to define `__class_getitem__`.
+# We're defining it here, instead of relying on django_stubs_ext, because
+# we don't want every user setting up django_stubs_ext just for this feature.
+# In theory, this can be replaced with `if TYPE_CHECKING` clause for base class,
+# but mypy does not support it at the time of writing.
```
last-partizan, line 27: These two lines are unrelated, and can be removed.

modeltranslation/admin.py
last-partizan: _get_form_or_formset is private API, and probably it's best to define type as just `Model`.

last-partizan:
Merged it! Thanks for your contribution!

mschoettle:
Thanks a lot for fixing the last issues and merging it!

Issue #748, opened 2024-07-17
modeltranslation 0.19.3+ with Django CMS 4.1.x - CMS Plugin functionalities broken


aacimov:
As stated in Django CMS issue tracker: https://github.com/django-cms/django-cms/issues/7948

Django CMS 4.1.x with modeltranslation active - custom CMS plugin attributes `allow_children`, and `child_classes` do not work correctly.

Each plugin that has `allow_children` has the ability to add child plugins but they are locked (one is not able to add plugins to children even though the `allow_children` with `child_classes` is set on Column in this example).

On the other hand, plugins that do not have explicitly set the `allow_children` attribute still have the ability to add children which is wrong.

**Image 1 - working example** with modeltranslation 0.19.2 - Text plugin has the **+ sign disabled** which is expected, and the Column plugins have the **+ sign enabled** which is expected:

<img width="398" alt="working 0 9 12" src="https://github.com/user-attachments/assets/560f1a2e-947b-4092-92e4-378fbf453d9c">

**Image 2 - not working example** with modeltranslation 0.19.3 and above - Text plugin has the **+ sign enabled** which is not expected, and the Column plugins have the **lock sign** which is not expected since no other plugins can be added into it:

<img width="398" alt="not_working 0 9 13" src="https://github.com/user-attachments/assets/92441d18-787f-40bc-a35a-894c17691a57">

## Project environment variables

- Django 4.2.13
- django_modeltranslation 0.19.2 (tried 0.19.3 and above which are not working, lower versions do work correctly)
- django-cms 4.1.1
- Python 3.10

aacimov:
Just tested with django-cms 3.11.x - affected as well.

last-partizan:
Hello.

Are you sure version 0.19.0 is working? because i there isn't much changes. But jump from 0.18 to 0.19 included some relatively big changes.

If you have time, please do additional testing using git version and `git bisect`.

aacimov:
@last-partizan yes, everything below 0.19.3 is working as expected. I will look into it over weekend probably and let you know.

benzkji:
for me as well, it is `django-modeltranslation<0.19.3` that works.

benzkji:
0.9.13 introduces changes on the admin clasess, that is what the cms uses, it makes sense to me.

last-partizan:
Yeah, there's some changes in admin, but they're mostly related to typing support.

If someone can provide repo with minimal reproduction, i can look into this.

[...]

Issue #745, opened 2024-07-05
Patching of class attribute getters causes error with modeltranslation 0.19.3+ and django-cms


benzkji:
I have this issue: https://github.com/django-cms/django-cms/issues/7948

I am not that familiar with admin internals. But to me, it looks like a basic call on an class level attribute causes a getter to be called, which in turn tries to access `self.admin_site`. I was not able to find that code in modeltranslation, though. But pinning <0.19.3 resolves the issue, whereas 0.19.3+ triggers. I was digging in the cms and modeltranslation code for quite a while, to get the infos provided here - but I am not able to find a solution currently. 

It is as well beyond my knowledge to say where this issue should be handled. cms oder here?

last-partizan:
Hello.

Please, provide full traceback and repo with minimal reproduction.

benzkji:
It is not that simple. You'll need to edit a html template in django-cms to see the attribute error - otherwise, django's template system will suppress the error. But yeah, I can try (override that template in my demo project...), it will help everybody understand that strange error.

benzkji:
@last-partizan I have tracked it down (probably), here is a PR with a test: #749

The dict access returns the class itself, which leads to the errors in django-cms (caused by how the template system accesses data, see https://github.com/django/django/blob/main/django/template/base.py#L886, where first, dict access is tried, and not getattr afterwards, if this is successful).

last-partizan:
Closing this as duplicate, let's move to the #748 as it has better reproduction example.

PR #749, opened 2024-08-07
provide tests for wrong dict access of admin class not raising TypeError

benzkji:
this is the cause for #745 and #748. I don't have a solution (yet).


last-partizan:
Thanks, closing this PR.

I will use your code as a base for a proper test and fix.

PR #752, opened 2024-08-08
feat: add changelog to project urls in package metadata

lociii:
Make it easier to find the change log on PyPI by listing it as one of the `Projekt-Links`.


last-partizan:
That's great, thank you!

PR #754, opened 2024-08-13
fix: Fix type error for Python 3.8

last-partizan:


mschoettle:
This brings back the issue that was fixed with #744. It seems that this was fixed for Python 3.8 which is not supported for this package anymore. Can we please revert this?

last-partizan:
Sure, created a PR reverting it.

After i merge it, please create separate PR adding a test for get_type_hits, so we accidentally do not break it again.

mschoettle:
Thanks! I'll add a test in a separate PR.

PR #755, opened 2024-08-13
django-cms fix

last-partizan:


last-partizan:
@aacimov @benzkji Please test if this fix is resolving your issues with django-cms.

benzkji:
thx alot, will check!

benzkji:
I installed with

`pip install -e git+https://github.com/deschler/django-modeltranslation.git@748-django-cms#egg=django-modeltranslation`

and the problem seems gone. Thanks a lot! @last-partizan

PR #756, opened 2024-08-20
fix(compat): Use `ForeignObjectRel.hidden` field from django 5.1

last-partizan:
Refs #719


Freakness109:
If you don't mind me adding the results of some investigations:
The main issue seems to be that `accessor_name` on `ForeignObjectRel` got turned into a cached property (https://github.com/django/django/commit/5e80390add100e0c7a1ac8e51739f94c5d706ea3, upstream https://code.djangoproject.com/ticket/35230). That means that as soon as `related_name` is overwritten in e.g. `fields.py`, the `accessor_name` will have an out-of-date string. Not sure how to fix it short of updating `accessor_name` manually.

last-partizan:
@Freakness109 your investigations was very useful!

Please, take a look at this PR, and test it with your project.

PR #758, opened 2024-09-04
test: add type hint test

mschoettle:
Follow-up to #754 and #757 to catch type hint errors in the future.

### DIFF
--- a/modeltranslation/tests/test_runtime_typing.py
+++ b/modeltranslation/tests/test_runtime_typing.py
@@ -1,4 +1,7 @@
+import sys
 from typing import get_type_hints
+
+import pytest
 
 from modeltranslation import admin, translator
 from modeltranslation.tests import models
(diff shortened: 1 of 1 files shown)


last-partizan:
@mschoettle that's interesting. But those files are using `from future import __annotations__`, so it's working for type-checkers but not for runtime?

If you don't care about python 3.9, then add `@pytest.mark.skip` based on python version.

modeltranslation/tests/test_runtime_typing.py
```diff
@@ -10,3 +12,9 @@ class TestInlineModelAdmin(
         admin.TranslationInlineModelAdmin[models.ForeignKeyModel, models.TestModel]
     ):
         pass
+
+
+def test_type_hints():
+    get_type_hints(translator.TranslationOptions)
+    get_type_hints(admin.TranslationAdmin)
+    get_type_hints(admin.TranslationInlineModelAdmin)
```
last-partizan, line 20:
```suggestion
@pytest.mark.parametrize("cls", [translator.TranslationOptions, admin.TranslationAdmin, admin.TranslationInlineModelAdmin])
def test_type_hints(cls):
    get_type_hints(cls)
```

mschoettle:
> @mschoettle that's interesting. But those files are using from future import __annotations__, so it's working for type-checkers but not for runtime?
>
> If you don't care about python 3.9, then add @pytest.mark.skip based on python version.

Yes, it is odd. Good idea, I added `@pytest.mark.skipif`.

Issue #766, opened 2024-11-13
Allow specifying a single string as fields


KommuSoft:
An error that probably occurs occasionally (like [here](https://stackoverflow.com/q/79185029/67579)) is that `fields` in a `TranslationOptions` is given a single string, not a list/tuple/... of strings.

This is a bit of a stupid mistake that we can probably prevent by just checking if the item is a string, and if that is the case wrap it in a tuple in the constructor, thus preventing the error from occurring?

Thoughts?

last-partizan:
Instead of converting it to a tuple, i raised the exception with a hint. Should give people instructions on what's wrong right away.

PR #767, opened 2024-11-13
wrap a fields in a sequence if it is a simple string fixing #766

KommuSoft:


modeltranslation/translator.py
```diff
@@ -77,7 +77,11 @@ class FieldsAggregationMetaClass(type):
     fields: Sequence[str]
 
     def __new__(cls, name: str, bases: tuple[type, ...], attrs: dict[str, Any]) -> type:
-        attrs["fields"] = set(attrs.get("fields", ()))
+        fields = attrs.get("fields", ())
+        if isinstance(fields, str):
+            attrs['fields'] = {fields}
```
last-partizan, line 82: This is bad idea, it's clearly an error, so maybe throw `ImproperlyConfigured` here?

Issue #800, opened 2025-11-12
Data migration for JSONField with default dict or list doesn’t populate to default language


manjurulhoque:
When a model field is defined as a JSONField with a default value of dict or list, running the command `python manage.py update_translation_fields` does not populate the existing data into the default language field as expected.

Example:

```
class MyModel(models.Model):
    json = models.JSONField(blank=True, null=True, default=dict)
```

After executing:

`python manage.py update_translation_fields`

The existing values in json are not copied or migrated to the corresponding default language field.

Expected behavior:
Existing data in JSONField should be populated into the default language field when running update_translation_fields.

Actual behavior:
The data remains empty or null for the default language field.

Environment:

Django version: 5.0.8
django-modeltranslation version: 0.19.5
Python version: 3.11

Additional context:
The issue seems related to how update_translation_fields handles fields with default values like dict or list.

PR #801, opened 2025-11-12
fix: fix update_translations_fields command to update default language field when JSON field has default dict or list #800

manjurulhoque:
It will solve the issue mentioned in https://github.com/deschler/django-modeltranslation/issues/800


last-partizan:
Please, add test-case for your issue and use `ruff format` to format code.

It should be similar to the existing `test_update_command_with_json_field`, but you need to add another field `json_with_defaults`.

Then, use `make recreate-migrations` to update test migrations.

haroon-ali-dev:
Tests added in #818.

manjurulhoque:
@haroon-ali-dev You are lucky person!

PR #814, opened 2026-03-10
fix: collect languages for `MainSwitch` from all modeltranslation fields

BeritJanssen:
I'm using django-modeltranslations in my project. In there, I have two models, `ChoiceList` and `Choice`:
```python 
class ChoiceList(models.Model):
    key = models.SlugField(max_length=64, primary_key=True)

class Choice(models.Model):
    key = models.SlugField(max_length=128)
    text = models.CharField()
    choicelist = models.ForeignKey(
        ChoiceList,
        on_delete=models.CASCADE,
        related_name='choices',
    )
```
Only the child model (`Choice`) has registered translation fields:
```python
@register(Choice)
class ChoiceTranslationOptions(TranslationOptions):
    fields = ('text',)
```

The admin form with this setup works really nicely already:
```python
class ChoiceInline(TranslationTabularInline):
    model = Choice

class ChoiceListAdmin(TabbedTranslationAdmin):
    model = ChoiceList
    inlines = [ChoiceInline]
```
BUT: the language `MainSwitch` on top of the form was empty, due to the way the `TranslationFieldGrouper` in `tabbed_translation_fields.js` was used for collecting these languages.

Screenshot 2026-03-09 at 15 44 34

This PR collects the languages to be displayed in the `MainSwitch` from all `.mt` classed elements and initializes the `TranslationFieldGrouper` with those languages.

Now the display is as it should be, and switching via the main switch works.
Screenshot 2026-03-10 at 11 45 34

### DIFF
--- a/modeltranslation/static/modeltranslation/js/tabbed_translation_fields.js
+++ b/modeltranslation/static/modeltranslation/js/tabbed_translation_fields.js
@@ -469,17 +469,9 @@ var google, django, gettext;
     var MainSwitch = {
       languages: [],
       $select: $("<select id='modeltranslation-main-switch' class='modeltranslation-switch'>"),
-
-      init: function (groupedTranslations, tabs) {
+      init: function (languages, tabs) {
         var self = this;
-        $.each(groupedTranslations, function (id, languages) {
-          $.each(languages, function (lang) {
-            if ($.inArray(lang, self.languages) < 0) {
-              self.languages.push(lang);
-            }
-          });
-        });
-        $.each(this.languages, function (idx, language) {
+        $.each(languages, function (idx, language) {
           self.$select.append(
             $(
               '<option value="' +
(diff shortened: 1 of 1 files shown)


modeltranslation/static/modeltranslation/js/tabbed_translation_fields.js
```diff
@@ -519,8 +512,9 @@ var google, django, gettext;
           .filter(":parents(.tabular)")
           .filter(":parents(.empty-form)"),
       });
+      var languages = getLanguages($(".mt").filter("input, textarea, select, iframe, div"));
```
last-partizan: Do we really need iframe and div here?

BeritJanssen: Probably not, will remove.

BeritJanssen:
Manual testing confirms that the selector doesn't have to be as broad as to include `iframe` and `div`.

modeltranslation/static/modeltranslation/js/tabbed_translation_fields.js
last-partizan: Why it was there in the first place? Did you use AI to write this?

BeritJanssen: No, I copied from another place where `.mt` elements were collected: lines 510 ff. (where the `TranslationFieldGrouper` is created for the change form). Not being familiar with the code base, I decided to collect all elements that might contain language codes as id/class name, since some use cases might use iframes or divs, apparently... I also didn't think collecting some irrelevant elements could do too much harm in terms of performance.

Issue #816, opened 2026-04-04
Apparently missing django-stubs-ext runtime dependency


oliverhaas:
Basically just title.

I some projects I get the error

`TypeError: type 'BaseModelAdmin' is not subscriptable`

and I think django-modeltranslations is just missing the runtime dependency of django-stubs, which is django-stubs-ext.

I guess since many projects are using django-stubs-ext due to other reasons, it's not always that obvious.

Will open a PR in a second.

last-partizan:
Well, that's interesting.

We have `modeltranslation/_typing.py` which doesn't requires django-stubs-ext (it's similar code).

We're importing monkeypatch() in the project `__init__.py`, but you still getting an error? Please show full traceback.

oliverhaas:
I missed `_typing.py` and that this is an intentional decision.

I just saw the above error in one of my fresh projects, saw "missing" django-stubs-ext in django-modeltranslation, and adding it worked as a fix for me.

I just thought I saw a small oversight late at night and wanted to flag it here :). 

Have a nice weekend.

PR #817, opened 2026-04-04
fix: add missing django-stubs-ext dependency

oliverhaas:
See #816

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,6 +14,7 @@ license = "BSD-3-Clause"
 requires-python = "<4,>=3.10"
 dependencies = [
     "Django>=5.1",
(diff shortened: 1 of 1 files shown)


last-partizan:
We aren't using django-stubs-ext in this project, and if you want to use monkeypatch() from that package, better to include it in your deps.

I hope we can resolve this issue with some other method.

oliverhaas:
> We aren't using django-stubs-ext in this project, and if you want to use monkeypatch() from that package, better to include it in your deps.
> 
> I hope we can resolve this issue with some other method.

Ok. The self-rolled monkeypatch of this package interferred with some of my project's typing.
Adjusted this on my end...

PR #821, opened 2026-09-03
Support dark mode and fix language selector in tabbed translation fields

vlazaroes:
### CSS (`tabbed_translation_fields.css`)

* **Dark mode support**: Now uses Django's admin CSS variables, each with a fallback to the original hardcoded value, so installations without these variables retain exactly the same appearance as before.
* **Layout with many languages**: The tab bar now wraps onto multiple rows as needed.
* **Consistent tab style**: Added `text-transform: uppercase` and `font-weight: 500` to the tab labels to follow the same visual conventions as the Django admin.

### JS (`tabbed_translation_fields.js`)

* **One-way select sync**: Changing the select moved the active tab, but clicking a tab never updated the select. `MainSwitch.update()` now also listens for jQuery UI's `tabsactivate` event to keep both in sync.
* **Tab tooltips**: Each tab now has a `title` attribute containing the language's full display name (e.g. hovering over "fr" shows "French").
* **`getLanguages()` bug with compound language codes**: It extracted the language code by splitting the field's `id` attribute and taking the last segment. This breaks for language codes that contain an underscore (e.g. `zh_cn` or `zh_tw`), causing the select to contain `"cn"`/`"tw"` instead of `"zh-cn"`/`"zh-tw"`.

Screenshot 2026-09-03 at 15 02 44
Screenshot 2026-09-03 at 15 02 52

### DIFF
--- a/modeltranslation/static/modeltranslation/css/tabbed_translation_fields.css
+++ b/modeltranslation/static/modeltranslation/css/tabbed_translation_fields.css
@@ -30,8 +30,8 @@ backward compatibility:
 .ui-tabs .ui-tabs-nav { margin: 0; padding: .2em .2em 0; }
 .ui-tabs .ui-tabs-nav li { list-style: none; float: left; position: relative; top: 1px; margin: 0 .2em 1px 0; border-bottom: 0 !important; padding: 0; white-space: nowrap; }
 .ui-tabs .ui-tabs-nav li a { float: left; padding: .5em 1em; text-decoration: none; }
-.ui-tabs .ui-tabs-nav li.ui-tabs-active, .ui-tabs .ui-tabs-nav li.ui-tabs-selected { margin-bottom: 0; padding-bottom: 1px; }
-.ui-tabs .ui-tabs-nav li.ui-tabs-active a, .ui-tabs .ui-tabs-nav li.ui-tabs-selected a, .ui-tabs .ui-tabs-nav li.ui-state-disabled a, .ui-tabs .ui-tabs-nav li.ui-state-processing a { cursor: text; }
+.ui-tabs .ui-tabs-nav li.ui-tabs-active, .ui-tabs .ui-tabs-nav li.ui-tabs-selected { margin-bottom: 0; padding-bottom: 0; }
+.ui-tabs .ui-tabs-nav li.ui-state-disabled a, .ui-tabs .ui-tabs-nav li.ui-state-processing a { cursor: text; }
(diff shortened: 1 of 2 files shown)


last-partizan:
That's nice, thank you.

In the future, please split different fixes into different PR's.

modeltranslation/static/modeltranslation/css/tabbed_translation_fields.css
```diff
@@ -41,14 +41,21 @@ backward compatibility:
...
 .ui-tabs { padding: 0; }
-.ui-tabs .ui-tabs-nav { padding: 5px 0 0 10px; border-bottom: 1px solid #EEEEEE; }
+.ui-tabs .ui-tabs-nav {
+    padding: 5px 0 0 10px;
+    border-bottom: 1px solid var(--hairline-color, #EEEEEE);
+    display: flex;
+    flex-wrap: wrap;
+    align-items: flex-end;
+    row-gap: 2px;
+}
```
last-partizan, line 51: `.ui-tabs-nav` is switched to `display:flex; align-items:flex-end`, but its `<li>` children (unchanged, further up the file) still carry `float:left; position:relative; top:1px`. Floats are inert inside a flex container, but `position:relative; top:1px` still applies, nudging every tab 1px below the flex baseline — a layout leftover from the float era that wasn't cleaned up when switching to flex.

---
generated by Claude Code

modeltranslation/static/modeltranslation/js/tabbed_translation_fields.js
```diff
@@ -542,12 +562,15 @@ var google, django, gettext;
     function getLanguages (mtFields) {
       let languages = [];
       $.each(mtFields, function (_idx, el) {
-          const ids = $(el).attr("id").split("_");
-          const lang = ids[ids.length - 1];
-          if ($.inArray(lang, languages) < 0) {
-            languages.push(lang);
+        $.each($(el).attr("class").split(" "), function (_idx, cls) {
```
last-partizan: `getLanguages` now does `$(el).attr("class").split(" ")` unconditionally. If an element in `mtFields` has no `class` attribute, `.attr("class")` returns `undefined` and `.split` throws — the old id-based lookup had no such dependency. Low likelihood since elements are pre-filtered to `.mt` fields, but easy to guard:

```js
$.each(($(el).attr("class") || "").split(" "), function (_idx, cls) {
```

---
generated by Claude Code

vlazaroes:
Fixed both :)

PR #822, opened 2026-09-07
fix: Rewrite F() expressions inside Q objects

MaxFreedomPollard:
`filter(visits=F("visits"))` rewrites both sides of the comparison to the active language, but `filter(Q(visits=F("visits")))` rewrites only the left side. The `Q` form compares a translation column against the untranslated original column, so it silently returns the wrong rows in any language whose value differs from the original field.

`MultilingualQuerySet._rewrite_filter_or_exclude` passes every keyword argument value through `_rewrite_f` (modeltranslation/manager.py:327), but `_rewrite_q` returned the value of a `Q` child unchanged (modeltranslation/manager.py:300): `return rewrite_lookup_key(self.model, q[0]), q[1]`. Only the lookup key was rewritten.

Concretely, for a row created as `visits_en=1, visits_de=2` with `de` active, master compiles `Q(visits=F("visits"))` to `WHERE "visits_de" = ("visits")` and matches nothing, while `filter(visits=F("visits"))` on the same row compiles to `WHERE "visits_de" = ("visits_de")` and matches.

The fix passes the value through `_rewrite_f`, which is the same call the keyword argument path already makes. That also routes it through the generic `get_source_expressions`/`set_source_expressions` handling in `_rewrite_f`, so nested expressions such as `F("visits") + 1` are rewritten as well.

I checked the common `Q` value shapes for regressions: `__in` with a list, with a queryset and with a `.values()` queryset, `=None`, `Value`, `Upper`, `Concat`, `Exists`, `Subquery`, `__range` and a negated `Q`. None of them crash or change results, since `_rewrite_f` leaves anything without `get_source_expressions` alone.

Verified on sqlite with Python 3.11.15 and Django 5.2.11:

```
$ uv run pytest -k test_q_with_f
1 passed, 196 deselected
# on unmodified master the same test fails with Q: (AND: ('visits_de', F(visits)))

$ uv run pytest
196 passed, 1 skipped

$ make lint
All checks passed!
42 files already formatted

$ make typecheck
Success: no issues found in 38 source files
```


last-partizan:
Thanks, merged!

